PR #6, opened 2024-05-17
refactor: the 'base' and a few 'header' components

orlinmalkja:


src/components/base/BaseIcon.vue
```diff
@@ -1,7 +1,12 @@
-<script setup>
+<script setup lang="ts">
 import { getIcon } from '@/utils/icons';
 
-const props = defineProps(['name']);
+//const props = defineProps(['name']);
+export interface Props {
+  name: string
```
orlinmalkja, line 6: Should I set the type of the name to be part of a list of names, or just let it string ?

src/components/header/GlobalHeader.vue
```diff
@@ -16,27 +16,28 @@
...
+}
+
+const props = withDefaults(defineProps<Props>(), {
+  configErrorTitle: ''
+})
 
-const show = computed(() => config.value?.header?.show);
-const manifests = computed(() => store.getters['contents/manifests']);
+const store = useStore();
 const config = computed(() => store.getters['config/config']);
```
orlinmalkja, line 37: Is it necessary to create an interface and assign it as a type to 'config' ?

src/components/base/BaseIcon.vue
paulpestov: string should be enough

src/components/base/BaseIcon.vue
```diff
@@ -1,7 +1,12 @@
-<script setup>
+<script setup lang="ts">
 import { getIcon } from '@/utils/icons';
 
-const props = defineProps(['name']);
+//const props = defineProps(['name']);
```
paulpestov, line 4: unused comment

paulpestov (approved):

PR #8, opened 2024-05-27
introduce 'config' Pinia store with other Vuex modules - drop the usage of 'config' Vuex module

orlinmalkja:
This MR addresses the following:

- create a 'config' Pinia store
- change other components to now use the 'config' Pinia store, instead of 'config' Vuex module
- remove unused code mainly in 'contents' Vuex module
- at the end of this MR: we will have a mixture of ['contents', 'annotations'] Vuex modules and 'config' pinia store


package.json
```diff
@@ -90,5 +90,8 @@
   ],
   "publishConfig": {
     "@subugoe:registry": "https://gitlab.gwdg.de/api/v4/projects/10921/packages/npm/"
+  },
+  "dependencies": {
```
paulpestov, line 94: Could you move that to devDependencies pls?

src/App.vue
```diff
@@ -26,17 +26,19 @@
...
 };
 </script>
 
-<script setup>
+<script setup lang="ts">
 import {
   computed, inject, onMounted, ref,
 } from 'vue';
 import { useStore } from 'vuex';
+import { useConfigStore } from './stores/config';
```
paulpestov, line 40: The path should use `@` instead of relative paths

src/components/header/PanelsToggle.vue
```diff
@@ -63,14 +63,15 @@ import { isMobile } from '@/utils/is-mobile';
...
 
 const store = useStore();
+const configStore = useConfigStore()
 const { t } = useI18n();
 
 const toggles = ref([]);
 const showDropdown = ref(false);
-
-const panels = computed(() => store.getters['config/config'].panels);
+const panels = computed(() => configStore.config.panels  );  //store.getters['config/config'].panels);
```
paulpestov, line 74: unused comment

src/types.d.ts
```diff
@@ -48,7 +48,52 @@ declare global {
...
+        secondary: string,
+        accent: string
+    }
+
+    interface ConnectorViewPanel {
+        id: number,
+        options?: Option_1_ConnectorViewPanel | Option_2_ConnectorViewPanel | Option_3_ConnectorViewPanel
+    }
+
+    type Option_1_ConnectorViewPanel = {
```
paulpestov, line 65: I think we should refine this a bit more. The types for different connector options should be more specific. A connector is in fact a real existing component that sits inside of a view (tab). For the sake of modularity we should push towards well-defined APIs (props/events) for each component. This means that we should define specifc types and interfaces for those views. For example a `MetadataView` component should be configured by a `MetadataViewConfigOptions` interface.

src/types.d.ts
```diff
@@ -179,6 +255,13 @@ declare global {
     }
     type TitleType = 'main' | 'sub';
 
+    interface ViewPanel {
```
paulpestov, line 258: Maybe we should call this `PanelView` or just `View` because it sounds like a special type of a `Panel` which is not the case.

package.json
orlinmalkja: yes, did it

src/App.vue
orlinmalkja: yes, changed the paths

src/types.d.ts
orlinmalkja: okay, renamed it to `View`

src/types.d.ts
orlinmalkja: I have worked in my previous two commits to address the refinement of `Config` interface.

src/components/header/PanelsToggle.vue
orlinmalkja: yes, removed it

src/types.d.ts
paulpestov: As discussed, please remove all non TextAPI Types and Interfaces for now and focus just on refactoring to Pinia store. We will deal with the correct typing of configs and their dependencies later.

orlinmalkja: @paulpestov yes, I just did it in my last commit:
Additionally in my last commit I did the following:
- removed the `Config Vuex Module` folder 
- used `Config Pinia Store` in some other components
- used only 'TextAPI' types in 'Config' Pinia store
- removed unused code

paulpestov (approved):

Issue #261, opened 2021-05-12
Roadmap about Implementation of the Variants in Annotations Panel


jfrer:
In GitLab by @krvoigt on May 12, 2021, 12:33

We already have wireframes and a concept. We need a roadmapping for the implementation steps.

Respective Ahiqar-ticket: [TIDO: implement concept for variants (AP2.b)](https://projects.academiccloud.de/projects/ahiqar/work_packages/18873/activity)

AC:
-  <del>a new view needs to be introduced "VariantsView" (this view should be configured as 3rd view in the annotation panel in Ahiqar config) </del>
- <del>variants list (variants selection):</del>
  - <del>displays a list of variant items (most likely normal annotation API items)</del>
  - <del>variant item:</del>
    - <del>displays an icon/text chip with an identifier of the witness, the given text content of the variant and a "dehighlight" button</del>
    - <del>clicking on the item should highlight/dehighlight the passage in the text panel with the same identifier icon displayed in the text</del>
    - a button to open the variants detail modal
- top bar (witnesses selection):
  - <del>at the top should be display a horizontal list of witness identifier icons that occur in the variant list below</del>
  - <del>clicking each icon should toggle (show/hide) all respective variants in the list below and highlight/dehighlight in the text panel </del>
  - there should be a button that toggles (show/hide) the witness description view 

- <del>selection modes:</del>
  -<del> there can be 2 selection modes - "all" and "single"</del>
  - <del>"all" is the behaviour described above</del>
  - <del>"single" mode does not display any variants in the list after initialization</del>
  - <del>the users can select and "collect" variants they are interested in from the text panel</del>

- variants detail modal:
  - <del>shows a column/table view with the selected variants from the variants list</del>
  - <del>shows the respective "original text" passage</del>

jfrer:
In GitLab by @krvoigt on May 12, 2021, 12:33

added to epic subugoe&40

jfrer:
In GitLab by @krvoigt on May 12, 2021, 12:37

marked this issue as related to #163

jfrer:
In GitLab by @paulpestov on May 6, 2024, 13:52

Here you can find a video explanation of the Wireframes rom #31: https://owncloud.gwdg.de/index.php/s/KHjfdFpGq1WEt87

jfrer:
In GitLab by @krvoigt on May 13, 2024, 12:20

[...]

PR #457, opened 2024-07-26
Finish 2. point of roadmap: variants list (variants selection)

orlinmalkja:


paulpestov (changes requested):
There are some things adresss here:

- Please refer to the video in the ticket #261 which shows a mock-up of the variants list and try to stick to the overall design idea. Improvement suggestions:
  - remove table-like headlines, align things inside an item to the left
  - smaller size of the "chip", more rounded, outlined without background
  - reproduce the same chip design when adding an element in the text panel
  - fix dark mode styles
- I am confused by the fact that we see 3 clickable variant items where on click you get 4 "pairs" selected at once. In the mock-up they always have only 1 witness identifier next to the entry. So if I assume right, the task should be to split up each of those 3 items into 4 individual items, so each witness/entry pair becomes clickable. The 4 "pairs" still refer to the same text panel element, though. So it is only possible to inspect only 1 witness/entry at a time in the text panel. But maybe there is still a general misunderstanding of the data @mathias-goebel.

paulpestov:
Also please chnge the icon names in the config and run `npm run fix` before submitting a PR. We need to bring that command to Github actions btw.

PR #460, opened 2024-08-11
Selecting/Deselecting an annotated text

orlinmalkja:
In this PR I address the following features:
- integrate together the functioning of selecting/deselecting a variant item in the Annotations panel and selecting/deselecting the annotated text in the Text Panel
- selecting an annotated text -> the witnesses 'chips' of that variant object should be shown in the text
- deselecting an annotated text -> remove the witnesses 'chips' from the text
- deal with the coloring of the witnesses chips

### DIFF
--- a/src/components/annotations/AnnotationVariantItem.vue
+++ b/src/components/annotations/AnnotationVariantItem.vue
@@ -160,7 +160,9 @@ function isOnlyThisVariantActive(witness) {
 }
 
 function pickVariantItemsSelection() {
-  // we pick the variant items selection, based on whether we have an active annotation or not
+  // we use this function in order to distinguish the state of variant items selection, before clicking on a certain variant item: either
+  //   a) no variant item is selected - initial state, variantItemsSelection copies the value of initVariantItemsSelection (every variant item has false value)
+  //   b) the variant item belongs to an active annotation - use the store computed property : 'activeAnnotSelectVariantItems'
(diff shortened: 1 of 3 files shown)


src/components/annotations/AnnotationVariantItem.vue
```diff
@@ -51,75 +56,142 @@ const props = withDefaults(defineProps<Props>(), {
...
     if (variantItemsSelection[wit] === true && wit!== witness) isOnlyThisVariantClicked = false
   })
   return isOnlyThisVariantClicked
 }
 
-function isVariantItemActive(witness): boolean{
-  return variantItemsSelection[witness] === true
-}
+function pickVariantItemsSelection() {
+  // we pick the variant items selection, based on whether we have an active annotation or not
```
paulpestov, line 163: Could you explain why we have to either of selctions, please? Actually we should describe why things happen in the code and not what happens, because from reading the code itself you can guess it.

src/components/annotations/AnnotationVariantItem.vue
```diff
@@ -51,75 +56,142 @@ const props = withDefaults(defineProps<Props>(), {
   toggle: () => null,
 })
 
-let variantItemsSelection = reactive({})
-let variantItemsColors = {}
-
+let initVariantItemsSelection = {}
```
paulpestov, line 59: It is strange that this variable has the same name as utils function. What is the job of this variable anyway? Can this be a prop?

paulpestov (approved):

PR #466, opened 2024-09-10
Tests variant item selection

orlinmalkja:
In this PR mainly I add the test about Variants Item Selection
I realised that the last commit we merged in main-variants : `fix: remove select All from VariantsView` had produced a bug in the simple selection of variant item, by not making it active when clicking it for the first time. In the second time it would become active 
Therefore I fixed it in this PR also.


orlinmalkja:
@paulpestov I made changes on the issues that we discussed. Could you please continue reviewing ?

paulpestov (approved):

PR #467, opened 2024-09-11
Tests deselect witnesses in drop down

orlinmalkja:


tests/cypress/e2e/variants.cy.js
```diff
@@ -74,4 +74,75 @@ const selectors = {
 
       })
     })
+
+    describe('Witnesses', () => {
+      it('Deselecting the witness remove the variant items of that witness and the witnesses in the transcription', () => {
```
paulpestov, line 79: Please give it a more "easy scannable" name.  As you can see the function is called "it", so you should go on with the sentence "it...deleselects a [first/second] witness from dropdown"

paulpestov (dismissed):

paulpestov (approved):

PR #472, opened 2024-09-17
Display all collectors' names and add a test for this

### LINKED ISSUE #471: Display all collectors instead of only the first one
Currently TIDO displays only the first array item from the Collector array. It should iterate over the array and display each item.

orlinmalkja:
Closes #471

### DIFF
--- a/src/components/metadata/CollectionMetadata.vue
+++ b/src/components/metadata/CollectionMetadata.vue
@@ -24,16 +24,12 @@ import MetadataItem from '@/components/metadata/MetadataItem.vue';
 
 const contentStore = useContentsStore();
 
-function getCollectorName(collection: Collection) : string[] | null {
+function getCollectorsName(collection: Collection) : string | null {
   if (!collection) return null;
   if (collection.collector.length === 0) return null;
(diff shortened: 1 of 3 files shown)


src/components/metadata/CollectionMetadata.vue
```diff
@@ -24,10 +24,16 @@ import MetadataItem from '@/components/metadata/MetadataItem.vue';
...
 
-function getCollectorName(collection: Collection) : string | null {
+function getCollectorName(collection: Collection) : string[] | null {
   if (!collection) return null;
   if (collection.collector.length === 0) return null;
-  return collection.collector[0].name;
+  if (collection.collector.length === 1) return [collection.collector[0].name];
+
+  let collectors: string[] = []
+  collection.collector.forEach((collector) => {
```
paulpestov, line 33: To simplify this, you could use `.map` function. Also the return type should stay the same (string). This can be achieved by using `.join(', ')`after `map` with an inline return. Please pay attention to the space after comma.

orlinmalkja:
thanks for your comment! I fixed it, so please continue with your review.

paulpestov (approved):

PR #476, opened 2024-09-19
Feat Nav buttons text configurable + add tests 

orlinmalkja:


index.html
```diff
@@ -27,7 +27,138 @@
 <script type="module" src="/src/main.js"></script>
 <script>
   window.addEventListener('load', function () {
-    window.tido = new window.Tido();
+    window.tido = new window.Tido({
```
paulpestov, line 30: Don't modify index.html pls

src/components/header/NavBar.vue
```diff
@@ -57,16 +57,13 @@ const hasNext = computed<boolean>(() => {
...
+
 const nextButtonLabel = computed<string>(() => (itemIndex.value === manifest.value.sequence.length - 1
-  ? `${t('next')} ${t(labels.value.manifest ? labels.value.manifest : 'Manuscript')}`
-  : `${t('next')} ${t(labels.value.item)}`));
+  ? `${t('next')} ${t('Manuscript')}`
+  : `${t('next')} ${t('item')}`));
 const prevButtonLabel = computed<string>(() => (itemIndex.value === 0
-  ? `${t('prev')} ${t(labels.value.manifest ? labels.value.manifest : 'Manuscript')}`
-  : `${t('prev')} ${t(labels.value.item)}`));
+  ? `${t('prev')} ${t('Manuscript')}`
```
paulpestov, line 65: Ideally, this looks like this `t('previous_manifest')`. No combined translations anymore. We talked about this, you remember?

orlinmalkja: yes you are right, we talked about this, but it seems that I had a misunderstanding, but now it's clear.

tests/cypress/e2e/header.cy.js
```diff
@@ -88,7 +88,7 @@ describe('Header - Item and Manifest changing', () => {
     cy
       .get(selectors.nextButton)
       .should('not.be.disabled')
-      .contains('Next Sheet');
+      .contains('Next Item');
```
paulpestov, line 91: This should actually stay as before. Ahiqar should now also change it's configration towards the new keys since there a manifest is called "Manuscript" and item is called "Sheet". In fact, Manuscript and Sheet were always Ahiqar-specific namings and had nothing to do with other projects.

examples/gfl-local.html
```diff
@@ -26,10 +26,7 @@
   window.addEventListener('load', function () {
     window.tido = new window.Tido({
       manifest: "http://localhost:8181/gfl/textapi/Z_1819-06-03_l/manifest.json",
-      labels: {
```
paulpestov, line 29: I think here is also a misunderstanding. The labels section should not be removed completely since it can have other dependencies. Did you check the metadata? I think it was used there. In this PR only new translation keys should be added.

index.html
orlinmalkja: okay, restored its content in a further commit

tests/cypress/e2e/header.cy.js
orlinmalkja: does that mean that we should create another 4 new keys only for Ahiqar, like 'Next Sheet', 'Previous Sheet', 'Next Manuscript' and 'Previous Manuscript' ?

examples/gfl-local.html
orlinmalkja: Okay, I understand, I have added a latest commit to address this.

orlinmalkja:
@paulpestov I have one question: This branch does not have the latest develop (I think for instance the part with showing all collectors is not in this branch). But still this branch does not have merge conflicts with develop. There shouldn't be any problem with merging right ? We are just adding the additions of this PR to the develop and not removing any of the develop's actual state right ?

I have added a latest commit to show the 'Sheet', 'Manuscript' only for Ahiqar editions. Feel free to continue reviewing.

examples/gfl-local.html
```diff
@@ -162,7 +162,14 @@
         }
       ],
       translations: {
-        Edition_license: 'Editionslizenz'
+        de: {
+          Edition_license: 'Editionslizenz',
+          next_page : "Nächste Seite",
```
paulpestov, line 167: Why did you add these keys here? TIDO should only have next_item, previous_item, next_manuscript, previous_manuscript. Every project should only configure those 4 keys.

src/components/header/NavBar.vue
```diff
@@ -57,16 +72,19 @@ const hasNext = computed<boolean>(() => {
...
 });
-const labels = computed<Labels>(() => configStore.config.labels || {
-  manifest: 'manifest',
-  item: 'item',
-});
-const nextButtonLabel = computed<string>(() => (itemIndex.value === manifest.value.sequence.length - 1
-  ? `${t('next')} ${t(labels.value.manifest ? labels.value.manifest : 'Manuscript')}`
-  : `${t('next')} ${t(labels.value.item)}`));
+
+// variables are named according to the general concept of ie (page = sheet, item);  (document = manuscript)
```
paulpestov, line 76: page and document are not according to general concept. They are only used in GFL. Instead we use item and manifest.

paulpestov:
> @paulpestov I have one question: This branch does not have the latest develop (I think for instance the part with showing all collectors is not in this branch). But still this branch does not have merge conflicts with develop. There shouldn't be any problem with merging right ? We are just adding the additions of this PR to the develop and not removing any of the develop's actual state right ?
> 
> I have added a latest commit to show the 'Sheet', 'Manuscript' only for Ahiqar editions. Feel free to continue reviewing.

It should work, lets see what happens

src/utils/translations.js
```diff
@@ -0,0 +1,13 @@
+
+export function areNavButtonsLabelsInConfig(config) {
```
paulpestov, line 2: This function is too specific. Indeed, you got the gist of the problem that we are facing when we start using keys from a translation config object: to merge the values with default config. Besides the fact that there is still a misunderstandig of the item/manifest terms usage across the whole PR, we can keep this way of identifying which translation to use for now. It just should not rely on counting of keys and so on since it's all dynamic. We cannot know how many keys project configs will provide. What we know for sure is the name of the key, this is the info that we should make use of.

orlinmalkja:
@paulpestov in my latest few commits I have tried to address your points. Please continue reviewing.

src/utils/translations.js
```diff
@@ -0,0 +1,31 @@
+
+import i18n from '@/i18n'
+
+// variables are named according to the general concept of ie (item = sheet, item);  (manifest = document, manuscript)
+const navButtonsDefaultTextArray = ['next_item', 'previous_item', 'next_manifest', 'previous_manifest']
+
+export function areNavButtonsLabelsInConfig(config) {
```
paulpestov, line 7: I think it should be fine fow now. What we want is actually a generic function `getTranslation` which looks up a translation string and tries to find it either in the user config or in the i18n files. Basically the `t` function is already something like that. It just doesn't check the user config. So maybe there is a way to just extend `t` or (worse) wrap `t` in a custom function and call it `$$t` or so.

examples/ahiqar-arabic-karshuni-local.html
```diff
@@ -31,6 +31,7 @@
       "item": "Sheet",
       "manifest": "Manuscript"
     },
+    "lang": "en",
```
paulpestov, line 34: Is this necessary to add? The default behaviour is that it switches automatically to english.

examples/ahiqar-arabic-karshuni-local-de.html
```diff
@@ -0,0 +1,159 @@
+<!DOCTYPE html>
```
paulpestov, line 1: Do we still need this?

orlinmalkja: no, we don't actually. We use glf-local file to test the german translations. Thus, I removed it

examples/ahiqar-arabic-karshuni-local.html
orlinmalkja: no, it's not, true

paulpestov (approved):

PR #479, opened 2024-10-01
test: add tests of single select mode for the variants feature

orlinmalkja:


orlinmalkja:
@paulpestov I am not sure why a few tests in github pipeline fail, while locally all tests pass with both ways: using cypress:headless and as a single test variants.cy.js (npm run cypress).

paulpestov (approved):

PR #484, opened 2024-10-14
Fix variants bugs #1

orlinmalkja:


src/components/annotations/variants/VariantsView.vue
```diff
@@ -61,10 +61,12 @@ onBeforeUnmount(() => unsubscribe())
 
 function allocateWitnessColorInVariantItem() {
   const colors = {}
-  annotationStore.witnesses.forEach((witness, i) => {
+  if(annotationStore.witnesses?.length > 0) {
```
paulpestov, line 64: Please use "early return" pattern, since wrapping an "if" around some code is kind of ugly

src/stores/annotations.ts
```diff
@@ -136,17 +136,20 @@ export const useAnnotationsStore = defineStore('annotations', () => {
...
-
+  
     filteredAnnotations.value
       .filter(annotation => !activeIds.includes(annotation.id))
       .forEach(annotation => {
         const selector = AnnotationUtils.generateTargetSelector(annotation);
         if (!selector) return;
         const selectorIsActive = activeIds.filter(id => selector === AnnotationUtils.generateTargetSelector(activeAnnotations.value[id])).length > 0;
-
-        if (!selectorIsActive && AnnotationUtils.getCurrentLevel(document.querySelector(selector)) < 0) {
```
paulpestov, line 147: Why do you remove the level check here?

src/stores/annotations.ts
```diff
@@ -209,13 +212,14 @@ export const useAnnotationsStore = defineStore('annotations', () => {
...
   const resetAnnotations = () => {
+    const textPanelHtml = document.querySelector('#text-content')
     if (annotations.value !== null) {
       annotations.value.forEach((annotation) => {
         const selector = AnnotationUtils.generateTargetSelector(annotation);
         if (selector) {
           AnnotationUtils.highlightTargets(selector, {level: -1});
           AnnotationUtils.removeIcon(annotation);
-          if (AnnotationUtils.isVariant(annotation)) {
+          if (AnnotationUtils.isVariant(annotation) && textPanelHtml.querySelector('.witnesses')) {
```
paulpestov, line 222: It is better to move this check into "removeWitness"

src/stores/annotations.ts
```diff
@@ -314,8 +318,8 @@ export const useAnnotationsStore = defineStore('annotations', () => {
       if (!target.dataset.annotation) {
         target = getNearestParentAnnotation(target);
       }
-
-      if (!target) {
+      
+      if (!target ||  target.getAttribute('data-annotation-level') < 0) {
```
paulpestov, line 322: We identified today that this wasn't really the issue of the bug. So this is not needed anymore I guess.

src/utils/annotations.js
```diff
@@ -241,27 +241,31 @@ export function removeIcon(annotation) {
...
-  else {
-    // get the target element and get the previous element - which we know is the 'witnesses' span list
-    // get the witnessesHtml element and append the witHtml element
+  else if(parentEl.children[indexOfTarget-1].classList.contains("witnesses")) {
     let witnessesHtmlEl = parentEl.children[indexOfTarget-1]
     witnessesHtmlEl.appendChild(witHtml)
   }
 }
 
+function createWitnessesHtml(witHtml, parentEl, targetHtmlEl) {
```
paulpestov, line 260: Please find better namings: the function receives an argument "witHtml" but its called "createWirnessesHtml", how can it be?

src/utils/annotations.js
```diff
@@ -241,27 +241,31 @@ export function removeIcon(annotation) {
...
+  const witnessesHtmlEl = document.createElement("span");
+  witnessesHtmlEl.classList.add('witnesses')
```
paulpestov, line 262: You can omit the words"witnessesHtml" if those are already in your function name (btw this applies also for classes or components). Call it "resultEl" or just "el", it will be fine.

src/stores/annotations.ts
orlinmalkja: Because there arises a bug with the target when it is null and it is not handled from the function getCurrentLevel(). Therefore I get first the target using document.querySelector() and then check if the target is null. If the target is null then no need to highlight the targets, therefore return;.

src/stores/annotations.ts
orlinmalkja: yes, true.

orlinmalkja:
For now, I am done with this PR. Please continue reviewing.

orlinmalkja:
@paulpestov from my side I am done with this PR. Please continue reviewing.

PR #486, opened 2024-10-15
Fix variants tab switching

paulpestov:


src/components/annotations/variants/VariantsView.vue
```diff
@@ -34,10 +36,16 @@ watch(
...
 
 const unsubscribe = TextEventBus.on('click', ({ target }) => {
-  const targetIsSelected = parseInt(target.getAttribute('data-annotation-level'), 10) > 0;
 
   const ids = getAnnotationIdsFromTarget(target)
 
+  // We check if the found annotation ids are currently displayed in the active tab, if not we skip the handling
+  const annotations = filteredAnnotations.value.filter((filtered) => ids.find(id => filtered.id === id))
+
+  if (annotations.length === 0) return
```
orlinmalkja, line 45: @paulpestov 
As a quick solution I would suggest: adding the part in bold fixes the current failing test in Single select mode
if (annotations.length === 0 && !annotationStore.isSingleSelectMode) return

orlinmalkja (approved):

PR #500, opened 2024-10-22
Introduce visible annotations in variants 

orlinmalkja:
use visibleAnnotations to show the current visible annotations, instead of filteredAnnotations.

Filtered annotations should contain the annotations for the current visible tab.
Visible annotations should contain only the current visible annotations, i.e when switching the single select mode on, it should be empty. When we click at a target and i.e 4 annotations are shown in the list, then visible annotations should contain only these annotations.


src/utils/annotations.js
```diff
@@ -250,7 +250,7 @@ export function addWitness(target, witness, color) {
   const witnessEl = createWitnessEl(witness, color)
   
   let isWrapper = false
-  if (parentEl.children.length === 1) {
+  if (targetIndex === 0) {
```
orlinmalkja, line 253: fix: remove array index bug when the target is at index 0 in its array

paulpestov:
Please add tests for switching between annotation tabs and the variant tab, also switching between different text panel tabs + variant tab. By introducing `visibleAnnotations` in the store, I cannot overview the consequences right now. I thought `visibleAnnotations` would be a local variable inside of VariantView on which the changes to `filteredAnnotations` would have been bypassed to. But now it is a whole new state management.

paulpestov:
I also see this when I open the arabic-karshuni page with the Dev endpoint:
Screenshot 2024-10-23 110255

orlinmalkja:
@paulpestov regarding your screenshot above: The witness value for a certain annotation is null, therefore it brings an error while setting a color to the variant item.

PR #506, opened 2024-10-24
Clicking a target should show the correct witnesses chips / select the right variant items by considering the witnesses drop down selection

orlinmalkja:
Bug: deselecting two witnesses from witnesses drop down and then clicking at the target, the number in the span 'x Variants selected'  is shown wrong. 

reason: clicking a target should show the correct witnesses chips / select the right variant items by considering the witnesses drop down selection

### DIFF
--- a/tests/cypress/e2e/variants.cy.js
+++ b/tests/cypress/e2e/variants.cy.js
@@ -376,6 +376,53 @@ const selectors = {
         .next()
         .should('not.have.class', 'active')
       })
+
+      it('should consider the witnesses drop down selection when clicking the target', () => {
+        // when clicking the target should show correct witnesses chips in text Panel and select the variant items according to witnesses drop down selection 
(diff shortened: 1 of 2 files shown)


orlinmalkja:
@paulpestov The last two commits are actually added in this PR. This branch is checkout out from the branch which deals with witness value null.

tests/cypress/e2e/variants.cy.js
```diff
@@ -376,6 +376,52 @@ const selectors = {
         .next()
         .should('not.have.class', 'active')
       })
+
+      it('should show correct witnesses chips in text Panel and number of selected variant Items when first deselecting a few witnesses and then clicking the target', () => {
```
paulpestov, line 380: Please find a shorter title

paulpestov (approved):

PR #517, opened 2024-11-05
fix: make the witnesses wrapper with flex-wrap css prop

orlinmalkja:


src/utils/annotations.js
```diff
@@ -272,7 +272,7 @@ export function addWitness(target, witness, color) {
 
 function createWitnessesWrapper() {
   const el = document.createElement("span");
-  el.classList.add('witnesses')
+  el.classList.add('witnesses', 't-flex', 't-flex-wrap')
```
paulpestov, line 275: I think t-inline-flex is better

paulpestov (approved):

PR #518, opened 2024-11-06
Bug in Variants apparatus: update variants list according to witnesses selection after disabling single select mode

orlinmalkja:


src/stores/annotations.ts
```diff
@@ -33,6 +35,10 @@ export const useAnnotationsStore = defineStore('annotations', () => {
     visibleAnnotations.value = payload
   }
 
+  function setActiveWitnessesIds(payload) {
+    activeWitnessesIds = payload
```
paulpestov, line 39: Does this work as expected? Arent you overwritng the reactive wrapper?

paulpestov (approved):

PR #520, opened 2024-11-07
fix: truncate and add ellipses for long witnesses

orlinmalkja:


src/components/annotations/variants/VariantItem.vue
```diff
@@ -8,9 +8,9 @@
     :data-annotation-id="annotation.id"
     @click="handleClick"
   >
-    <div class="t-w-4/12 t-flex">
+    <div class="t-w-1/3 t-flex">
       <div
-        class="t-relative t-rounded-3xl t-box-border t-px-2 t-py-1 t-text-xs t-flex-grow-0 t-flex t-items-center t-font-semibold t-whitespace-nowrap"
+        class="t-relative t-truncate t-w-7/8 t-rounded-3xl t-box-border t-px-2 t-py-1 t-text-xs t-flex-grow-0 t-font-semibold"
```
paulpestov, line 13: Could you also add a title attribute with the full name, so it will be shown on hover?

paulpestov (approved):

Issue #470, opened 2024-09-12
Add share button with citation hint to TIDO


mokkalicious:
### Description:
This issue was noticed in Ahiqar project. --> https://projects.academiccloud.de/wp/42467
In TIDO Simon suggested a share button with the citation hint.

This is about the citation hint within the TIDO Metadata tab as seen in this example:
![Image](https://github.com/user-attachments/assets/c216f768-d1f8-4215-8fc2-7053ae2db7dc)

### Acceptance Criteria

Share button is located in position which makes sense UX wise

MVP: Share button copies citation hint to clipboard

_Nice2Have: button opens dialogue to choose between options like, e.g.

copy to clipboard

open in your email software

open in Zotero(?)_

paulpestov:
How should we identify that it's a citation? If I remember correctly that part is a dynamic key/value pair, so projects might use a different key for their citations like "Zitation" or "Zitat" or "How to Cite". I see two options:
- Make it configurable as metadata panel option, e.g. citationKey = "Citation"
- Make it a separate TextAPI key, like we have it with "actor"
@mathias-goebel @mweidling @mokkalicious

mathias-goebel:
to generalize this, a "share" button might become a "copy" button and can be available to all fields. to keep the UI clean, a button might become available on hover, since a user who wants to copy the citation statement will always move a cursor to the respective field

paulpestov:
Not sure about displaying the copy button everywhere. What about values that contain dynamic HTML or Markdown (that decision that we still have to make)? We currently have a PR for the citation key solution in the config.

orlinmalkja:
Hi guys
I am not sure either of displaying the copy button in every metadata item. 

in the current PR #530 a solution to this issue is done according to 'Make it configurable as metadata panel option, e.g. citationKey = "Citation"'. 
The copy button is shown only in the 'row' when there is a config value of citation the same as TextApi metadata key 'Citation'.

paulpestov:
The discussion that goes beyond the citation key should be in a separate issue.

PR #530, opened 2024-11-21
Share citation copy button

orlinmalkja:
Please check the functionality with this url: https://api.ahiqar.sub.uni-goettingen.de/textapi/ahiqar/syriac/collection.json


paulpestov:
There is a discussion about this feature here #470.

paulpestov:
The design of the button should be more seamless. Check out the implementation of it here in GitHub https://github.com/subugoe/tido/pull/530/files. If you look at the file path at the top of the list you will discover the small copy icon. Upon click it turns into a small check icon.

orlinmalkja:
Okay, I have another question: Do we still show the confirmation message 'Copied successfully' below the button ?

paulpestov:
I'd say yes

paulpestov:
The tooltip should be positioned absolutely. It should not move othr things in the view. Also the copy icon should be placed inline behind the text.

src/utils/icons/icons.js
```diff
@@ -1,31 +1,61 @@
...
+export const dropdown =
+  '<svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 16 16"><path fill="currentColor" fill-rule="evenodd" d="M1.646 4.646a.5.5 0 0 1 .708 0L8 10.293l5.646-5.647a.5.5 0 0 1 .708.708l-6 6a.5.5 0 0 1-.708 0l-6-6a.5.5 0 0 1 0-.708"/></svg>';
+export const copy =
+  '<svg class="w-6 h-6 text-gray-800 dark:text-white" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="currentColor" viewBox="0 0 24 24"><path fill-rule="evenodd" d="M18 3a2 2 0 0 1 2 2v10a2 2 0 0 1-2 2h-1V9a4 4 0 0 0-4-4h-3a1.99 1.99 0 0 0-1 .267V5a2 2 0 0 1 2-2h7Z" clip-rule="evenodd"/><path fill-rule="evenodd" d="M8 7.054V11H4.2a2 2 0 0 1 .281-.432l2.46-2.87A2 2 0 0 1 8 7.054ZM10 7v4a2 2 0 0 1-2 2H4v6a2 2 0 0 0 2 2h7a2 2 0 0 0 2-2V9a2 2 0 0 0-2-2h-3Z" clip-rule="evenodd"/></svg>';
```
paulpestov, line 60: Why is everything reformatted? Also please don't add tailwind classes here. Use them on the outer element. Also look at the other icon svg. They have 1em as width/height. This is for using font-size on the outer element to control the size of the icon.

orlinmalkja: I am not sure, why everything is shown as reformatted. I actually only added the extra copy icon. Anyway in one of my latest commits I restored the icons as they are in develop branch and then just added my icon.

paulpestov: Please check my comment above, the width/height should be 1em.

paulpestov:
I still have a few points: 

- The icon should sit vertically centered, right now its slightly above the text line
- When you switch to the checkmark icon, the text line should not jump, I think this is due to a different icon size
- The "tooltip" text should sit at the bottom center of the icon and not at bottom left of the text line (again, just chck the copy button from Github) 
- The icon feels too heavy, maybe we should try some outlined icon like this: https://icones.js.org/collection/all?s=copy&icon=mingcute:copy-2-line (you can just copy the SVG)

src/components/metadata/MetadataItem.vue
```diff
@@ -25,31 +29,62 @@
...
+    (panel) => panel.label.toLowerCase() === "metadata"
+  );
+  let metadataView;
+  if (panelMetadata.length > 0) {
+    // when metadata is one separate panel
+    metadataView = panelMetadata[0].views[0];
+  } else {
+    // when there is one panel containing content and metadata views
+    const panelContainingMetadata = panels.filter((panel) =>
+      panel.label.toLowerCase().includes("metadata")
```
paulpestov, line 66: This line breaks TIDO when testing with EUPT config. The reason is that you try to parse a label value. Please find a better way. We should never parse user given values in generic products!

paulpestov:
The points I stated previously are done 👍

src/components/metadata/MetadataItem.vue
orlinmalkja: @paulpestov thank you for the hint! I selected the metadata view by conditionally filtering the views using the  connector id

orlinmalkja:
@paulpestov I think I fixed the selection part of the metadata view and centered the tooltip below the share button!

paulpestov (approved):

PR #533, opened 2024-11-29
TIDO v5: show panels wrapper which contains text and switch between different types of text content

orlinmalkja:


orlinmalkja:
@paulpestov Shall we display by default the first text content type or should one make the initial display configurable ? 
I think it is not necessary to make it configurable, but just show initially the first text content type, it seems natural to me. The reason against making it configurable, is that obtaining the desired content type can be achieved by a very simple action - clicking at another button.

paulpestov:
Somehow I get this error at `npm run dev`:

  primeicons/primeicons.css (imported by .../tido/src/App.tsx)
  primereact/button (imported by .../tido/src/components/panel/TextTypes.tsx)

src/main.tsx
```diff
@@ -1,9 +1,9 @@
-import { StrictMode } from 'react'
-import { createRoot } from 'react-dom/client'
-import App from './App.tsx'
+import { StrictMode } from "react";
```
paulpestov, line 1: I suggest to keep the whole app without semicolons and with single quotes instead of double quotes. It just looks cleaner in my opinion. For this there are ESLint rules that we can add in the config. If you then apply the "lint on save" feature in your IDE it would automatically pick up the new rules.

tsconfig.json
```diff
@@ -15,12 +14,17 @@
...
+
     /* Linting */
     "strict": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true
   },
-  "include": ["src"]
+  "include": ["./"]
```
paulpestov, line 29: Why this change?

src/components/panel/Panel.tsx
```diff
@@ -0,0 +1,97 @@
...
+// prop: url - should be the url of collection or manifest
+const Panel: FC = ({ url }) => {
+  const { config, setConfig } = useContext(ConfigContext);
+  const [data, setData] = useState();
+  const [itemUrl, setItemUrl] = useState("");
+  const [text, setText] = useState<React.ReactNode | undefined>();
+  const [textTypes, setTextTypes] = useState([]);
+  const [activeText, setActiveText] = useState("");
+
+  async function getItemUrl(documentData): string {
```
paulpestov, line 20: Type?

src/components/panel/Panel.tsx
```diff
@@ -0,0 +1,97 @@
...
+    if (itemData["content"].length === 0) return;
+
+    const content = itemData["content"];
+    for (let i = 0; i < content.length; i++) {
+      types.push(getContentType(content[i].type));
+    }
+    setTextTypes(() => types);
+  }
+
+  function getContentType(value): string {
```
paulpestov, line 46: Type?

orlinmalkja:
> Somehow I get this error at `npm run dev`:
> 
> Error: The following dependencies are imported but could not be resolved: primeicons/primeicons.css (imported by .../tido/src/App.tsx) primereact/button (imported by .../tido/src/components/panel/TextTypes.tsx)

have you installed primereact locally ?

paulpestov:
I think its missing in package.json

tsconfig.json
orlinmalkja: restored it back.

src/components/panel/Panel.tsx
orlinmalkja: added the type

src/components/panel/Panel.tsx
orlinmalkja: added the type

paulpestov:
I had errors during the installation again, this time it was the SASS compiler. In any case, I think we don't need SASS anymore since it was mainly used for some global styling that we can solve otherwise. Also the nesting approach is becoming a native CSSS standard (https://caniuse.com/css-nesting). I'm already making use of it in the `preflight.css`.

src/App.tsx
```diff
@@ -1,9 +1,34 @@
-function App() {
+import PanelsWrapper from './components/PanelsWrapper';
+import { useEffect, useState } from 'react';
+import { useConfig} from '@/contexts/ConfigContext'
+
+
+
+function App({customConfig}) {
```
paulpestov, line 7: Type?

src/App.tsx
```diff
@@ -1,9 +1,34 @@
...
+
+  const [loading, setLoading] = useState(true)
+  const { config, updateConfig} = useConfig()
```
paulpestov, line 10: updateConfig not exported, code style

src/contexts/ConfigContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+  );
+};
+
+function useConfig(): ConfigProvider {
+  return useContext(ConfigContext)
+}
+
+
+
+export { ConfigProvider, useConfig };
```
paulpestov, line 39: The export gives me a fast fresh ESLint error. Since the official React [example](https://codesandbox.io/p/sandbox/react-dev-59fz7l?file=%2Fsrc%2FTasksContext.js&utm_medium=sandpack) shows to use it in one file I suggest to add supress the `react-refresh/only-export-components` ESlint rule.

src/contexts/ConfigContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+const ConfigContext = createContext({});
+
+interface ConfigProvider {
+  config?: Config,
+  setConfig?: () => void,
+  openedPanels?: Panel[],
+  setOpenedPanels?: (panels: Panel[]) => void
+}
+
+interface ConfigProviderProps {
```
paulpestov, line 12: `..Props {` Here it shows an irregular whitespace as `NBSP` for me

src/App.tsx
```diff
@@ -1,9 +1,34 @@
...
+  useEffect(() => {
+    // load Config
+    // TODO: url config will need to get read
+    function loadConfig() {
+      const totalConfig = {...config, ...customConfig}
+      updateConfig(totalConfig)
+      setLoading(false)
+    }
+    loadConfig()
+  }, [])
```
paulpestov, line 21: Shouldn't the dependencies array contain all variables it is using inside?

src/components/PanelsWrapper.tsx
```diff
@@ -0,0 +1,36 @@
+import { FC, useEffect } from 'react';
+import Panel from '@/components/panel/Panel';
```
paulpestov, line 2: I think this is imported accidentally. Instead it should come from types.d.ts

src/components/PanelsWrapper.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { useConfig } from '@/contexts/ConfigContext';
+import { getManifestUrl, getCollectionUrl } from '@/utils/panel';
+
+const PanelsWrapper: FC = ({}) => {
+  const { config, openedPanels, setOpenedPanels } = useConfig();
+
+  function initOpenedPanels(panels: Panel[]) {
```
paulpestov, line 9: Is this wrapper function needed? You could just call `setOpenedPanels` directly.

src/types.d.ts
```diff
@@ -178,6 +183,10 @@ declare global {
         info: string,
         warning: string
     }
+    interface Panel {
```
paulpestov, line 186: We should rename this to `PanelConfig` because we have a component with the same name.

src/components/PanelsWrapper.tsx
```diff
@@ -0,0 +1,36 @@
...
+    }
+    initOpenedPanels(config.panels);
+  }, []);
+
+  
+  const panels = openedPanels ?
+    openedPanels.length > 0 &&
+    openedPanels.map((panel: Panel, i: number) => (
+      <div key={i} className="t-mr-6">
+        <Panel url={ panel.collection ? getCollectionUrl(panel): getManifestUrl(panel)} />
```
paulpestov, line 28: Just pass the whole `PanelConfig` as prop here and let the component figure out what it should load. This way we can also read other upcoming config options inside it.

src/components/panel/Panel.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const [error, setError] = useState<boolean | string>(false)
+  const [loading, setLoading] = useState<boolean>(true)
+
+
+  async function getItemUrl(documentData: Manifest | Collection): Promise<string | null> {
+    // if collection - then we should read the api data from the manifest and get its first sequence item id
+    // if manifest - we retrieve the first sequence item id
+
+    // panel in the config having a document with the prop url
+    const panel: Panel | undefined = getPanel(url, config)
```
paulpestov, line 32: This should become obsolete as soon as you pass the PanelConfig as prop.

src/App.tsx
orlinmalkja: I think updateConfig is not necessary, since setConfig itself in the ConfigContext can receive a new Config as an argument and update the config. Therefore, I removed updateConfig. This should still be fine, or ?

src/App.tsx
orlinmalkja: converted the component to FC component and used a typescript interface for its type.

src/contexts/ConfigContext.tsx
orlinmalkja: In my code eslint does not raise any error.. Is it still the case for you ?

src/App.tsx
orlinmalkja: I am not sure, when have a config change which affects only a specific component, then we would rerender the App component and its children, right ?  

Moreover, the loadConfig in App.tsx will be executed at the very first beginning.

src/components/PanelsWrapper.tsx
orlinmalkja: Actually it is meant to import the Panel component and not the PanelConfig (I updated the name in types.d.ts for the Panel config).

src/components/PanelsWrapper.tsx
orlinmalkja: yes, true. I removed initOpenedPanels()

src/types.d.ts
orlinmalkja: yes, sounds good, I renamed it.

src/utils/http.tsx
```diff
@@ -0,0 +1,21 @@
+
+export async function get(url: string): Promise<Manifest | Collection | null> {
+  let response: Manifest | Collection | null = null
```
paulpestov, line 3: This is not generic I'm afraid. Mentioning specific interfaces and using specific error messages should be avoided.

src/config.tsx
```diff
@@ -0,0 +1,15 @@
+const defaultConfig: Config = {
+  globalTree: true,
+  panels: [
+    {
+      entrypoint:
```
paulpestov, line 5: Hm I thought you would implement this in some next PR. This was just an idea, I'm not sure about this yet.

index.html
```diff
@@ -27,8 +27,19 @@
...
   window.addEventListener('load', function () {
-    window.tido = new window.Tido();
-  });
+    window.tido = new window.Tido({
+      globalTree: true,
+      panels: [
+        {
+          entrypoint:
+            'https://api.ahiqar.sub.uni-goettingen.de/textapi/ahiqar/syriac/collection.json',
+          m: 3,
```
paulpestov, line 36: Hm I thought you would implement this in some next PR. This was just an idea, I'm not sure about this yet.

src/App.tsx
orlinmalkja: Actually I think that useEffect can make it possible to execute something only once. If I place code outside of useEffect it will be executed repeatedly. 

I removed useEffect() from PanelsWrapper, but to my opinion it seems like it is necessary in other components.

src/components/PanelsWrapper.tsx
orlinmalkja: Yes it is a good idea to pass the whole PanelConfig as a prop. I did it.

src/utils/http.tsx
orlinmalkja: Yes, in my latest commit I try to create a generic function to fetch api data. On that commit I try to parse the data based on the content type which I get from the headers of the response received. Regarding the content types of the response I had a look on this [article](https://medium.com/@anatoliiyatsenko/understanding-fetch-api-response-methods-and-the-content-type-header-6dcbe7b24ded)

src/App.tsx
paulpestov: Hm I still see the useEffect in PanelsWrapper. Also why do you need useEffects in other components?

paulpestov (approved):

src/main.tsx
```diff
@@ -5,8 +5,14 @@ import './css/style.css'
 
 import App from './App.tsx'
 
-window.Tido = function Tido(config = {}) {
-    createRoot(document.getElementById('app')!).render(<App customConfig={config} />)
+declare global {
```
orlinmalkja, line 8: @paulpestov I merged the PR, but I have still one question. Which is the idea behind the lines 8-12 in this file ?

paulpestov: Just to satisfy Typescript

PR #543, opened 2025-01-06
Switch text views - Use OpenSeaDragon for the image

orlinmalkja:
This PR aims to fulfill the following requirements:

- Use a Zustand store to store the content of the current visible panels. The data for one panel is stored as a key-value pair. 
- Switch between text views when clicking the respective icon. 
- Use Open SeaDragon for the image and add the image action buttons
- Open a popover on click


tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "isolatedModules": true,
     "moduleDetection": "force",
     "noEmit": true,
-    "jsx": "react-jsx",
+    "jsx": "preserve",
```
paulpestov, line 15: Why this change?

index.html
```diff
@@ -30,10 +30,22 @@
...
       globalTree: true,
       panels: [
-        {
+        { 
           entrypoint: {
             url: 'https://api.ahiqar.sub.uni-goettingen.de/textapi/ahiqar/syriac/collection.json',
             type: 'collection'
+          },
+          colors: {
+            primary: '#1E40AF'
```
paulpestov, line 39: I don't think we should allow to set the color per panel. This should be a global setting.

src/components/panel/Panel.tsx
```diff
@@ -1,155 +1,33 @@
...
   }
 
-  if (loading) {
-    return <div> Loading the data of this panel...</div>
-  }
-
   return (
-    <div className="panel t-flex t-flex-col t-w-[600px] t-ml-[6%] t-border-solid t-border-2 t-border-slate-200 t-rounded-lg t-mt-4 t-px-2.5 t-pt-8 t-pb-6">
+    <div className="panel t-flex t-flex-col t-w-[600px] t-mr-6 t-border-solid t-border-2 t-border-slate-200 t-rounded-lg t-mt-4 t-px-2.5 t-pt-8 t-pb-6">
+      <PanelTopBar panelId={panelId} />
```
paulpestov, line 24: After looking closely at this code, we can see that the panelId is being passed in every child component individually. I don't think that this is a good pattern. Maybe we can make use of the Context Provider by just keeping the panelId there and letting the components get it. This way we can get rid of all panelId props.

src/components/panel/central-content/PanelCentralContent.tsx
```diff
@@ -0,0 +1,72 @@
+import { FC, useEffect, useState } from 'react'
```
paulpestov, line 1: Just call the folder "views" instead of "central-content"

src/store/ContentStore.tsx
```diff
@@ -0,0 +1,51 @@
+import { create } from 'zustand'
+
+interface VisiblePanels {
+  [id: string]: PanelContentStore
+}
+
+interface ContentStoreTypes {
+    openedPanels: VisiblePanels // or panels: each panel has one opened item
```
paulpestov, line 8: Just call it "panels"

src/store/ContentStore.tsx
```diff
@@ -0,0 +1,51 @@
+import { create } from 'zustand'
+
+interface VisiblePanels {
```
paulpestov, line 3: Just call it "PanelStates"

src/types.d.ts
```diff
@@ -142,6 +142,16 @@ declare global {
         modules?: Module[]
     }
 
+    interface PanelContentStore {
```
paulpestov, line 145: I think this should be called "PanelState"

src/types.d.ts
```diff
@@ -142,6 +142,16 @@ declare global {
...
+        item: Item,
+        texts?: string[],
+        contentTypes?: string[],
+        contentIndex: number,
+        textViewIndex: number,
```
paulpestov, line 150: viewIndex?

src/components/PanelsWrapper.tsx
```diff
@@ -1,21 +1,146 @@
...
+          setError('The document type for this entrypoint url is not valid')
+          continue
+        }
+
+        const itemData = await readData(panels[i], documentType) 
+        if (!itemData) continue
+
+        const contentTypes: string[] = getContentTypes(itemData.content)
+        // id of PanelContent object
+        const panelId = 'panel-' + i
```
paulpestov, line 107: We don't gain anything by using a key value object in the store when our keys are still depending on loop indexes. Please use uuid or something here.

paulpestov (approved):

PR #553, opened 2025-01-20
Data store

paulpestov:
This PR tries to resolve some issues with data loading and panel rendering that I noticed in our current state. 

First of all it introduces a DataStore for dealing with collections. For now just keep collections there but we can imagine to extend it's responsibilities to keep other data as well. With this store we gain a global state of loaded collections that is accessible for any component. 

Second, this PR refactors the way panels are rendered. In the previously version the PanelWrapper would load everything. The problem was that the PanelWrapper component became to long as it was looping though the config and pre-loaded every piece of data for the panels. Now it just takes the config and passes the relevant part (a single panel config) to a Panel component. This is the only loop. Once a Panel is instantiated, it will handle everything itself. The idea is to let the Panel generate everything it needs for it to be rendered: panelId, collection, manifest, item. Having the DataStore we can easily access that information. Later we could check whether that collection was already loaded and return that data instead of request the API again.


src/components/panel/Panel.tsx
```diff
@@ -1,27 +1,75 @@
...
+        const manifest = await apiRequest<Manifest>(collection.sequence[config.manifestIndex ?? 0].id)
+        const item = await apiRequest<Item>(manifest.sequence[config.itemIndex ?? 0].id)
+        const contentTypes: string[] = getContentTypes(item.content)
+
+        addPanelContent(panelId, {
+          item,
+          contentIndex: 0,
+          viewIndex: 0,
+          contentTypes,
+          primaryColor: 'blue',
```
orlinmalkja, line 42: Primary color should not be attached to a panel, but be global right ?

src/components/panel/Panel.tsx
```diff
@@ -1,27 +1,75 @@
...
+  const [error, setError] = useState<boolean | string>(false)
+  const [loading, setLoading] = useState(true)
+  const [panelId, setPanelId] = useState<string | undefined>(undefined)
+
+  useEffect(() => {
+    const panelId = crypto.randomUUID()
+    const collectionUrl = config.entrypoint.url
+    const init = async () => {
+      try {
+        setLoading(true)
```
orlinmalkja, line 31: I think this is unnecessary, since its default state is already true.

src/store/DataStore.tsx
```diff
@@ -0,0 +1,18 @@
+import { create } from 'zustand'
+import { apiRequest } from '@/utils/api.ts'
+
+interface DataStoreType {
+  collections: Collection[]
+  initCollection: (url: string) => Promise<Collection>
+}
+
+export const dataStore = create<DataStoreType>((set, get) => ({
+  collections: [],
```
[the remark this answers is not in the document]
paulpestov: 1. To keep the collection response objects here
2. I don't mind to be it key-value
3. What do you mean by minimizing? This store is totally separated from the panels.

src/components/panel/Panel.tsx
paulpestov: Yes, no idea why it is there

src/components/panel/Panel.tsx
paulpestov: Ah ok, the default state then should be false.

orlinmalkja:
The PR seems good to me!

orlinmalkja (approved):

PR #559, opened 2025-01-22
Dynamic collections

orlinmalkja:
This PR aims to accomplish the following goals:

- Open a new modal when clicking `Add` Button 
- Create the Tree View of the currently opened panels and show it in the Modal
- Add a new panel dynamically when either entering a collection url or clicking at a tree item
- Showing an error message when the user has not provided a way to open a new collection dynamically

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,8 @@ import PanelsWrapper from './components/PanelsWrapper'
 import { FC } from 'react'
 import { configStore } from '@/store/ConfigStore.tsx'
 
+import TopBar from '@/components/TopBar'
+
 interface AppProps {
   customConfig: Config
 }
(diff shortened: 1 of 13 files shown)


paulpestov (commented):
The structure you chose to render the tree is not very flexible. In your code the components are not very reusable. They rather follow a strict hierarchy. Imagine we will implement later the lazy loading feature that would be required on manifest and items level both. YOu would need to implement it twice because you have sepcial components for manifest and item sub trees. 

A better approach would be general tree node components that can render recursively other tree node comps as children. Each node component can toggle it's children display if there are any. And later you just add the lazy loading function onto the children within that node component. 

Our idea of having collections in collections would also come closer.

PR #562, opened 2025-01-30
Dynamic collections: Creation of a new panel by using local tree modal

orlinmalkja:
This PR aims to accomplish the following goals:

- Open a new (local) modal when clicking Add Button
- Create the tree of the provided collections and show in the modal
- Add a new panel dynamically when either entering a collection url or selecting a tree item


src/utils/tree.ts
```diff
@@ -0,0 +1,92 @@
...
+    manifestIndex: number
+}
+
+interface ItemIndices {
+    collectionUrl: string,
+    manifestIndex: number,
+    itemIndex: number
+}
+
+export async function createTree(panels: PanelConfig[]) {
```
paulpestov, line 15: As a slight improvement, this function could receive an array of collection URLs instead of dealing with panelConfig.

src/store/DataStore.tsx
```diff
@@ -6,17 +6,38 @@ interface CollectionMap {
...
+  },
+
+  async getCollection(collectionUrl: string): Promise<Collection> {
+    if (collectionUrl in get().collections) return get().collections[collectionUrl]
+
+    const collection = await get().initCollection(collectionUrl)
+    return collection
+  },
+
+  updateTreeNodes: (newNodes: TreeNode[]) => {
```
paulpestov, line 39: This does the same as initTreeNodes. Use only one function: setTreeNodes

src/contexts/TreeContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+  onSelect(node: TreeNode): void,
+
+  onExpand(node: TreeNode): void,
+
+  onCollapse(node: TreeNode): void,
+}
+
+const TreeProvider: FC<TreeProviderProps> = ({ children, onSelect, onExpand, onCollapse }) => {
+
+  function onClick(node: TreeNode) {
```
paulpestov, line 27: I'm afraid there is still a misconception about what should happen on click. I know I pointed out "onExpand" and "onCollapse" being callbacks from outside. This is still true but they after looking at this they seem to be less important for this stage of the Tree implementation. 

What should actually happen here is this:
`if ('leaf' in node) select(node); else if (!node.expanded) expand(); else if (node.expanded) collapse();`

and 

`function select(node: TreeNode) { /* internal behaviour */ onSelect(node); }` and similar for the others.

Do you see the missing piece? It's the selection, exand, collapse itself. The `onSomething` callback should be just the reaction upon that behaviour after the actual thing happened, as if we provide a hook into our component for the users. The actual behaviour should happen within the component. 

So what we actually want is to call `getChildren` at the right time. I suggest to remove `onExpand` and `onCollapse` from the Tree props and add `getChildren`prop instead. So ContentModal only defines **how** to get children and not **when**. 

We could also keep it for now, but you will notice for the global tree that you would copy the onExpand and onCollapse functions in the sidebar again and again. Maybe it will be more clear, then.

src/store/DataStore.tsx
orlinmalkja: yes, true

src/store/ConfigStore.tsx
```diff
@@ -1,13 +1,23 @@
 import { create } from 'zustand'
 
+
```
paulpestov, line 3: ?

src/components/tree-modal/ContentModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+      updatedTree[collectionIndex].expanded = true
+
+    } else if (type === 'manifest') {
+      const { collectionIndex, manifestIndex } = getManifestIndices(node.key)
+      const manifestNode = { ...updatedTree[collectionIndex].children[manifestIndex] }
+
+      if (!('children' in manifestNode)) manifestNode['children'] = await getChildren(node)
+
+      updatedTree[collectionIndex].children[manifestIndex] = { ...manifestNode }
```
paulpestov, line 103: There are some Typescript issues here

src/store/DataStore.tsx
```diff
@@ -1,22 +1,35 @@
...
+  treeNodes: TreeNode[],
   initCollection: (url: string) => Promise<Collection>
+  setTreeNodes: (newTreeNodes: TreeNode[]) => void,
+  getCollection: (collectionUrl: string) => Promise<Collection>,
 }
 
 export const dataStore = create<DataStoreType>((set, get) => ({
   collections: {},
+  treeNodes: [],
+  clickedItemUrl: '',
```
paulpestov, line 16: Is this still used? If yes it shouldn't be here.

src/utils/http.ts
```diff
@@ -2,40 +2,51 @@ const BLOB_CONTENT_TYPES = ['application/pdf', 'image/png', 'image/jpeg', 'audio
...
-    )
-  }
 }
 
-function createErrorResponse(message: string = '', code: number = 500): ErrorResponse {
-  return { success: false,
-    message,
-    code,
-  }
+export function isValidUrl(value: string) {
```
paulpestov, line 44: If not used please remove

src/components/LocalTreeModal.tsx
```diff
@@ -0,0 +1,29 @@
...
+const LocalTreeModal: FC<LocalTreeProps> = ({ TriggerButton }) => {
+
+  // TODO: add a [loading, setLoading] => which shows the pop over when the tree has been loaded -> TreeView Component updates the loading of its parent
+
+  // function of this component: create a new Panel
+
+
+  return <div className="local-tree-modal">
+    <Popover>
+      <PopoverTrigger className="open-tree-button t-h-8 t-w-10 t-relative">
```
paulpestov, line 19: Are these styles required? everything should be handled in the Popover comps actually.

src/components/LocalTreeModal.tsx
```diff
@@ -0,0 +1,29 @@
...
+        {TriggerButton}
+      </PopoverTrigger>
+      <PopoverContent className="t-bg-white t-absolute t-z-10">
```
paulpestov, line 22: Same here

src/components/LocalTreeModal.tsx
```diff
@@ -0,0 +1,29 @@
+import { FC, ReactNode } from 'react'
+
+import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover'
+import ContentModal from '@/components/tree-modal/ContentModal'
+
+interface LocalTreeProps {
+  TriggerButton: ReactNode
+}
+
+const LocalTreeModal: FC<LocalTreeProps> = ({ TriggerButton }) => {
```
paulpestov, line 10: Naming is not good. Why Local? Maybe call it TreeSelectionModal?

src/components/tree-modal/ContentModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { dataStore } from '@/store/DataStore'
+
+
+import TreeView from '@/components/Tree.tsx'
+import InputField from '@/components/base/InputField.tsx'
+import { ClosePopover } from '@/components/ui/popover'
+import { createTree, getItemIndices, getManifestIndices, getChildren } from '@/utils/tree'
+
+
+const ContentModal: FC = () => {
```
paulpestov, line 13: You mean ModalContent? Better TreeSelelectionModalContent

src/components/Tree.tsx
```diff
@@ -0,0 +1,42 @@
...
+  nodes: TreeNode[],
+
+  onSelect(node: TreeNode): void,
+
+  onExpand(node: TreeNode): void,
+
+  onCollapse(node: TreeNode): void,
+}
+
+// ({ nodes, onSelect, onExpand })
```
paulpestov, line 18: ?

src/components/Tree.tsx
```diff
@@ -0,0 +1,42 @@
...
+const Tree: FC<TreeProps> = ({ nodes, onSelect, onExpand, onCollapse }) => {
+
+
+  const tree =
+    nodes.length > 0 &&
+    nodes.map((collection, i) => (
+      <div
+        key={i}
+        className=""
```
paulpestov, line 27: ?

src/components/Tree.tsx
```diff
@@ -0,0 +1,42 @@
...
+      >
+        <TreeNode node={collection}/>
+      </div>
+    ))
+
+
+  return <div className="tree t-h-96 t-overflow-hidden t-overflow-y-auto">
+    <TreeProvider onSelect={onSelect} onExpand={onExpand} onCollapse={onCollapse}
+    >
```
paulpestov, line 36: ?

src/contexts/TreeContext.tsx
```diff
@@ -0,0 +1,50 @@
+import { ReactNode, createContext, useContext, FC } from 'react'
+
+const TreeContext = createContext<TreeType | undefined>(undefined)
+
+interface TreeType {
+  onClick(node: TreeNode): void,
+
```
paulpestov, line 7: Remove the blank lines, remove the commas

src/contexts/TreeContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+const TreeProvider: FC<TreeProviderProps> = ({ children, onSelect, onExpand, onCollapse }) => {
+
+  function onClick(node: TreeNode) {
+    if ('leaf' in node) onSelect(node)
+    else if (!node.expanded) onExpand(node)
+    else if (node.expanded) onCollapse(node)
+  }
+
+  return (
+    <TreeContext.Provider value={{ onClick, onSelect, onExpand, onCollapse }}>
```
paulpestov, line 34: remove onClick

src/components/tree/TreeNode.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const TreeNode: FC<TreeNodeProps> = ({ node }) => {
+
+  const { onClick } = useTree()
+
+  const itemRef = useRef(null)
+
+  function handleNodeClick() {
+    onClick(node)
+    // add class t-bg-primary
```
paulpestov, line 18: ?

src/contexts/TreeContext.tsx
orlinmalkja: thank you for your suggestion. I will undertake changes on the tree props, functions at another PR.

src/components/LocalTreeModal.tsx
orlinmalkja: no, they are not needed, removed them

src/components/LocalTreeModal.tsx
orlinmalkja: yes, removed them

src/components/LocalTreeModal.tsx
orlinmalkja: local was meant for the local tree - its not the global tree, yes, right, it sounds better

src/contexts/TreeContext.tsx
orlinmalkja: it happens only on this file, I don't know why

src/contexts/TreeContext.tsx
orlinmalkja: i would like to keep it, so that the tree node can use it

orlinmalkja:
I have added new commits to address your raised points. Please continue reviewing!

src/contexts/TreeContext.tsx
paulpestov: resolve it in your IDE later, but commit it the right way into repo first ;)

src/contexts/TreeContext.tsx
paulpestov: somehow it is shown as unused for me

paulpestov (approved):

PR #563, opened 2025-02-05
Add the feature of Global Tree 

orlinmalkja:
- In this PR we offer to the user the option to update or add a new Panel using a Global Tree. 
- The implementer of TIDO can utilize Global Tree functionality by specifying in the config: `globalTree` as `false` or `true`.


orlinmalkja:
@paulpestov I made other small changes I intended to. Please continue reviewing!

src/App.tsx
```diff
@@ -1,21 +1,45 @@
...
 
-const App: FC<AppProps> = ({ customConfig }) => {
+const App: FC<AppProps> = ({ children, customConfig }) => {
   const addCustomConfig = configStore((state) => state.addCustomConfig)
   addCustomConfig(customConfig)
 
+  const collections = dataStore(state => state.collections)
+  const setTreeNodes = dataStore(state => state.setTreeNodes)
+
+  const [showGlobalTree, setShowGlobalTree] = useState(false)
```
orlinmalkja, line 23: When toggling the global tree display, I want to communicate this change from TopBar Component to the Global tree Component. In order to do this, I thought to keep this state 'showGlobalTree' in their common nearest parent component - in this case it's the `App`.

To prevent unnecessary re-rendering in `PanelsWrapper` when this state changes, I added this component as a child to the `App` in main.stx.

paulpestov: This is even worse since it's leaking the app structure to main.tsx. Please, keep App and main tsx files clean. Just use some boolean on the panelStore.

orlinmalkja: Okay, thanks, i made the changes

paulpestov (approved):

PR #567, opened 2025-02-12
feat: add sync panel scrolling feature - example TextAPI data, target…

paulpestov:
New features: 

- new button in top bar to open a "sync" modal 
- in the modal all current panels can be chosen to be in "sync" mode
- once you click confirm in the modal, the "sync" will be enabled and a yellow frame appears around respective panels
- with the yellow frame there are 2 arrow buttons at the top center of each panel
- with enabling "sync" mode all "sync" targets will be highlighted in the texts of each panel with grey background
- also each "sync" panel scrolls in parallel 
- clicking on a "sync" target in each panel with cause the other "sync" panels to scroll to the position of their equivanet target (click target 1 in panel 1 -> scroll to target 1 in panel 2)
- clicking the arrows will cause the same as above but it remembers the active target index, so it can navigate through the target in the panel
- selecting new "sync" panels in the modal will reset all listeners appended previously and reapply them for the new selected "sync" panels

misc:
- refactoring tree files for a better file structure
- not using "Modal" comp for global stree selection but rather a custom fake modal because the trigger button and positioning should be handled by shadcn popover

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,9 @@
   },
   "private": false,
   "devDependencies": {
+    "@eslint/js": "^9.16.0",
+    "@radix-ui/react-popover": "^1.1.4",
+    "@radix-ui/react-slot": "^1.1.1",
     "@types/openseadragon": "^3.0.10",
     "@types/react": "^18.3.12",
     "@types/react-dom": "^18.3.1",
@@ -91,10 +94,5 @@
   },
   "files": [
     "dist"
-  ],
-  "dependencies": {
-    "@eslint/js": "^9.16.0",
-    "@radix-ui/react-popover": "^1.1.4",
-    "@radix-ui/react-slot": "^1.1.1"
-  }
+  ]
 }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -35,7 +35,7 @@ const App: FC<AppProps> = ({ customConfig }) => {
   return (
     <div className="tido t-flex t-flex-col">
       <TopBar/>
-      <div className="t-flex">
+      <div className="t-flex t-overflow-hidden">
         <GlobalTree/>
         <PanelsWrapper/>
       </div>
--- a/src/components/SelectParallelPanels.tsx
+++ b/src/components/SelectParallelPanels.tsx
@@ -39,8 +39,9 @@ const SelectParallelPanels: FC = () => {
   }
 
   function unsync() {
-    setSelected({})
-    update(Object.keys(selected))
+    const newSelected = {}
+    setSelected(newSelected)
+    update(Object.keys(newSelected))
   }
 
   return <>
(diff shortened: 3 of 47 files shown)


orlinmalkja:
There is a bug when trying to open a new collection with this URL: https://api.dev.ahiqar.sub.uni-goettingen.de/textapi/ahiqar/arabic-karshuni/collection.json 

1. Steps to reproduce:
   - click New Button in Top Ba,  add the url 'https://api.dev.ahiqar.sub.uni-goettingen.de/textapi/ahiqar/arabic-karshuni/collection.json ' and then click select
2. Expected result: open a new panel with this collection url
3. Obtained result: screen becomes empty 
image

package.json
```diff
@@ -36,6 +36,9 @@
   },
   "private": false,
   "devDependencies": {
+    "@eslint/js": "^9.16.0",
+    "@radix-ui/react-popover": "^1.1.4",
```
orlinmalkja, line 40: which is the reason that you installed this library ?

package.json
```diff
@@ -36,6 +36,9 @@
...
+    "@radix-ui/react-slot": "^1.1.1",
```
orlinmalkja, line 41: the same question for this library

src/App.tsx
```diff
@@ -35,7 +35,7 @@ const App: FC<AppProps> = ({ customConfig }) => {
   return (
     <div className="tido t-flex t-flex-col">
       <TopBar/>
-      <div className="t-flex">
+      <div className="t-flex t-overflow-hidden">
```
orlinmalkja, line 38: do you know why adding this style makes it possible to view the whole text in the panel

src/components/SelectParallelPanels.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Button } from '@/components/ui/button.tsx'
+import { panelStore } from '@/store/PanelStore.tsx'
+import { scrollStore } from '@/store/ScrollStore.tsx'
+
+const SelectParallelPanels: FC = () => {
+  const panelStates = panelStore(state => state.panels)
+  const update = scrollStore(state => state.update)
+  const scrollPanelIds = scrollStore(state => state.panelIds)
+
+  const [selected, setSelected] = useState<{ [key: string]: boolean }>(scrollPanelIds.reduce((acc, cur) => {
```
orlinmalkja, line 11: Which is the aim to set the initial value of this state using a `reducer` function ?

paulpestov: The the modal opens, we need to read the initial state of currently selected sync panels. This will set the buttons as selected.

package.json
paulpestov: this was actually added when i installed the popover, however i also don't understand how this was not added before. Usually this is added by executing shadcn commands.

src/App.tsx
paulpestov: this prevents any overflow in this container, especially vertically, so the panels can handle scrolling themselves, was that your question?

orlinmalkja (approved):

PR #569, opened 2025-02-13
Fix annotation icons

paulpestov:
annotation icons in text have now primary color and rendered as svg elements and not as content prop


src/utils/annotations.js
```diff
@@ -205,6 +145,7 @@ export function handleRangeSelector(selector) {
 }
 
 export function addIcon(element, annotation, iconName) {
+  console.log('add')
```
orlinmalkja, line 148: console.log

orlinmalkja (approved):

PR #571, opened 2025-02-16
Panel header

paulpestov:


src/components/panel/PanelTitle.tsx
```diff
@@ -0,0 +1,22 @@
+import { FC } from 'react'
+import { usePanel } from '@/contexts/PanelContext.tsx'
+
+const PanelHeader: FC = () => {
```
orlinmalkja, line 4: you meant `PanelTitle` right ?

paulpestov: no its actually the whole header

orlinmalkja: but the content is only the title of the Panel. The `PanelHeader` you have defined in another file.

PR #572, opened 2025-02-21
feat: add "fitPanels" config option, control panel widths with flex values

paulpestov:
Adding a new config key "fitPanels" at the rool level of config. This boolean controls whether all the configured panels should fit in the container or whether they should keep a certain width (defaultWidth) and make the container overflow. If fitPanels set to true, the "width" key inside each panelConfig behaves differently than with fitPanels = false. 

Test with:
1 Panel - fit:true - null - Desktop
1 Panel - fit:true - 0.5 - Desktop
1 Panel - fit:true - 2 - Desktop
1 Panel - fit:false - null - Desktop
1 Panel - fit:false - 0.5 - Desktop
1 Panel - fit:false - 2 - Desktop
1 Panel - fit:true - null - Mobile
1 Panel - fit:false - 2 - Mobile
2 Panels - fit:true - null 2 - Desktop
2 Panels - fit:true - null 2 - Mobile
4 Panels- fit:true - 0.5 0.5 1 2 - Desktop
4 Panels- fit:true - 0.5 0.5 2 2 - Desktop
4 Panels- fit:true - 0.5 0.5 2 2 - Desktop - toggle different panel show
4 Panels- fit:true - null null null 2 - Desktop
4 Panels- fit:false - 0.5 0.5 2 2 - Desktop
4 Panels- fit:false - null null null 2 - Desktop
4 Panels- fit:false - null null null 2 - Desktop - toggle different panel show
7 Panels- fit:true - null null null 2 null null null - Desktop
7 Panels- fit:false - null null null 2 null null null - Desktop
7 Panels- fit:false - 0.5 5 null 2 null null null - Desktop
4 Panels- fit:true - 0.5 0.5 1 2 - Mobile
4 Panels- fit:true - null null null 2 - Mobile
4 Panels- fit:false - null null null 2 - Mobile
7 Panels- fit:false - 0.5 5 null 2 null null null - Mobile

### DIFF
--- a/src/components/panels/Panel.vue
+++ b/src/components/panels/Panel.vue
@@ -1,8 +1,8 @@
 <template>
   <div
-    class="panel t-flex-shrink-0 t-flex t-flex-col t-overflow-hidden t-rounded-lg t-bg-gray-50 dark:t-bg-gray-800
+    class="panel t-flex t-flex-col t-overflow-hidden t-rounded-lg t-bg-gray-50 dark:t-bg-gray-800
     t-border dark:t-border-gray-700 t-max-h-screen md:t-h-auto"
-    :style="{width: `${width}px`}"
+    :style="{ flexBasis, flexGrow, flexShrink }"
   >
     <div class="panel-header t-py-3 t-px-4 t-flex t-justify-between t-items-center">
       <div class="caption t-font-bold">
@@ -127,7 +127,7 @@
 
 <script lang="ts">
 import {
-  computed, nextTick, ref, watch,
+  computed, nextTick, onMounted, onUnmounted, ref, watch,
 } from 'vue';
 import { useI18n } from 'vue-i18n';
 import TabView from 'primevue/tabview';
@@ -176,22 +176,34 @@ export default {
       default: () => { },
     },
     activeView: Number,
-    defaultWidth: Number,
   },
   setup(props, { emit }) {
-    const { isMobile } = useResize();
     const configStore = useConfigStore();
     const contentStore = useContentsStore();
     const { t } = useI18n();
+    const { isMobile, onResize } = useResize();
 
+    const defaultWidth = 300;
     const tabs = ref([]);
     const activeTabIndex = ref(0);
     const unsubscribe = ref(null);
     const isLoading = ref(false);
-
-    const width = computed( () => props.defaultWidth * getWidth(props.panel.width));
+    const flexGrow = ref(1);
+    const flexShrink = ref(1);
+    const flexBasis = ref('0%');
 
     const item = computed<Item>(() => contentStore.item);
+    const config = computed(() => configStore.config);
+
+    let unsubscribeResize;
+
+    onMounted(() => {
+      unsubscribeResize = onResize(() => updateFlexValues(config.value.fitPanels, props.panel.width, isMobile.value));
+    });
+
+    onUnmounted(() => {
+      if (unsubscribeResize) unsubscribeResize();
+    });
 
     watch(
       () => props.activeView,
@@ -210,14 +222,9 @@ export default {
 
     watch(item,() => init(props.panel.views));
 
-    function getWidth(configValue) {
-      if (!configValue || typeof configValue !== 'number') return 1;
-      if (isMobile.value || configValue < 1) return 1;
-      if (configValue > 10) return 10;
-      return configValue;
-    }
-
     function init(views) {
+      updateFlexValues(config.value.fitPanels, props.panel.width, isMobile.value)
+
       tabs.value = [];
       if (unsubscribe.value !== null) unsubscribe.value();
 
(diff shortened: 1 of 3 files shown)


orlinmalkja:
If I understand that right: when the `fitPanels` is false and the `widths` of 4 panels are given: `1.5`, `1.5` , `3`, `3`. Then the width size of each panel would be the `widthMultiplier x default Width: `450px`, `450px`, `900`, `900`, right ?
If yes the we should have a horizontal overflow of the panels, which is not the case. The panels fit the container, by splitting proportionally the container based on proportion values

orlinmalkja:
When fitPanels is set to false, then the panels overflow horizontally if the summation of (width Multiplier for each panel with the default width) exceeds the container width, right ?

src/components/panels/Panel.vue
```diff
@@ -243,6 +250,25 @@ export default {
       });
     }
 
+    function updateFlexValues(fitPanels: boolean, widthMultiplier: number, isMobile: boolean) {
```
orlinmalkja, line 253: it would be helpful if a few lines would describe the functionality of this function. It is not obvious to me what is happening here.

orlinmalkja:
> If I understand that right: when the `fitPanels` is false and the `widths` of 4 panels are given: 1.5, 1.5 , 3, 3. Then the width size of each panel would be the `widthMultiplier x default Width: 450px, 450px, 900, 900, right ? If yes the we should have a horizontal overflow of the panels, which is not the case. The panels fit the container, splitting proportionally the container based on proportion values

Well I found the reason why this doesn't work: The `**fitPanels` in customConfig** is not validated correctly. I will add a suggestion in the respective code. With the new validation then the we obtain the expected behaviour.

src/stores/config.ts
```diff
@@ -279,6 +280,10 @@ import { i18n } from '@/i18n';
       return isValid;
     }
 
+    function validateFitPanels(value) {
+      return !!(value);
```
orlinmalkja, line 284: when the value is false, then this function returns false. Actually it needs to return true, even if the value is false. I would rewrite this function as 
`function validateFitPanels(value) {
      if (typeof value === 'boolean') return true;
      return false
}`

The line which calls this function is 406, namely `...(validateFitPanels(fitPanels) && { fitPanels })`. This line adds the fitPanels, only if the function returns true.

orlinmalkja:
@paulpestov These are the results from your test cases. I think the results are reasonable and correct.

-   1 Panel - fit:true - null - Desktop.  - result: the panel fits the container  
-   1 Panel - fit:true - 0.5 - Desktop    - result: the panel fits the half of the container 
-   1 Panel - fit:true - 2 - Desktop.      - result: the panel fits the container
-   1 Panel - fit:false - null - Desktop. - result: panel fits the container
-   1 Panel - fit:false - 0.5 - Desktop.  - result: panel fits the container
-   1 Panel - fit:false - 2 - Desktop.      - result panel fits the container 
-   1 Panel - fit:true - null - Mobile.      - result: the panel fits the container
-   1 Panel - fit:false - 2 - Mobile.        - result: panel has twice x default width
-   2 Panels - fit:true - null 2 - Desktop. - result: both panels fit the container, the second panel has twice the width of the first one
-   2 Panels - fit:true - null 2 - Mobile.       - result: both panels fit the container, the second panel has twice the width of the first one
-   4 Panels- fit:true - 0.5 0.5 1 2 - Desktop - result: all panels have the right proportional widths and the panels fit the container
-   4 Panels- fit:true - 0.5 0.5 2 2 - Desktop - result: all panels have the right proportional widths and they fit the container
-   4 Panels- fit:true - 0.5 0.5 2 2 - Desktop - toggle different panel show -   result: all panels have the right proportional widths and they fit the container
-   4 Panels- fit:true - null null null 2 - Desktop - result: all panels have the right proportional widths and they fit the container
-   4 Panels- fit:false - 0.5 0.5 2 2 - Desktop - each panel has width x default width
-   4 Panels- fit:false - null null null 2 - Desktop. result: each panel width: multipl x default width + overflow
-   4 Panels- fit:false - null null null 2 - Desktop - toggle different panel show - same
-   7 Panels- fit:true - null null null 2 null null null - Desktop. - 7 panels fit in the container, their width is proportionally set according to width multiplier
-  7 Panels- fit:false - null null null 2 null null null - Desktop - each panel width: multipl x default width + overflow
-  7 Panels- fit:false - 0.5 5 null 2 null null null - Desktop - each panel width: multipl x default width + overflow
-  4 Panels- fit:true - 0.5 0.5 1 2 - Mobile - all panels fit the container + proportional widths
-  4 Panels- fit:true - null null null 2 - Mobile - all panels fit the container + proportional widths
-  4 Panels- fit:false - null null null 2 - Mobile - each panel has width x multiplier
-  7 Panels- fit:false - 0.5 5 null 2 null null null - Mobile - each panel width: multipl x default width + overflow

src/components/panels/Panel.vue
paulpestov: Instead of calculating the width in pixels for the panels I chose to use flex values. This is due to the self-managing behavior of flex items, especially when we toggle panels. At each toggle we have to recalculate and this didn't end up well. In flexbox the items that grow take their need amount of space. 

With that said, the important switch is the flex-basis value of "0%" vs. "a calculated width". With "0%" it just relies on the grow and shrink values which we need if we want to fit in the container.  Otherwise we just reuse flex-basis instead of width together with the old way of calculating the width config key for the exact pixel values, but with a possibility to grow if the size is smaller than the container.

paulpestov:
Thank you very much for testing!

src/stores/config.ts
paulpestov: fixed

orlinmalkja (approved):

PR #588, opened 2025-03-27
fix: display annotations when no refs are given

paulpestov:


orlinmalkja:
when `refs` in annotationPage is [], then the annotations are not shown due to an error: 
image

You can reproduce the error by adding the following line `refs = []` before [this line](https://github.com/subugoe/tido/blob/91613d840be5646841731490d745449ef1deb38a/src/stores/annotations.ts#L201)

paulpestov:
I assume you are trying to see variants where refs is []. In that case I would see this as a valid error which hints you that your API is doing something wrong. (I know it's not that clear the console error above but it's not worth IMO to improve the error messages for v4).

orlinmalkja (approved):

orlinmalkja:
@paulpestov yes that I was trying to do. Alright, I approved the PR.

PR #595, opened 2025-04-01
Fix loading issues

paulpestov:


src/components/panel/views/PanelCentralContent.tsx
```diff
@@ -24,6 +26,12 @@ const PanelCentralContent: FC = () => {
 
   const [error, setError] = useState<string | null>(null)
 
+  function getContentUrlByType(type: string | undefined) {
+    console.log(type)
```
orlinmalkja, line 30: console log

src/components/panel/views/PanelCentralContent.tsx
```diff
@@ -35,7 +43,9 @@ const PanelCentralContent: FC = () => {
       }
     }
 
-    const contentUrl = panelState?.item?.content[0].url ?? null
+    if (!panelState?.contentTypes.length) return
+    const contentUrl = getContentUrlByType(panelState?.contentTypes[activeContentTypeIndex].toLowerCase())
+    console.log(contentUrl)
```
orlinmalkja, line 48: console log

paulpestov:
removed

orlinmalkja:
I have a question: if the 'content Type' of a certain text is not provided correctly, we don't show it at all in Contents Type Toggle buttons. As a consequence we do not show the text of that Type, since we switch only between the content types. Is it supposed to be like that ? Would the alternative of showing the text even without having a correct content type value be reasonable ?

PR #609, opened 2025-04-09
Merge configs

paulpestov:


src/main.tsx
```diff
@@ -21,7 +22,7 @@ window.Tido = function Tido(config = {} as Config) {
   }
 
   const style = document.createElement('style')
-  style.innerHTML = `${container || '#app'} {${getRGBColor(theme?.primaryColor ?? '#3456aa', 'primary')}}`
+  style.innerHTML = `${container || '#app'} {${getRGBColor(theme?.primaryColor ?? defaultConfig.theme?.primaryColor, 'primary')}}`
```
orlinmalkja, line 25: when primaryColor in user config has i.e an int value, then an error arises inside getRGBColor(). So the primaryColor value should be validated before being given as input to getRGBColor()

orlinmalkja (approved):

PR #611, opened 2025-04-10
Translations feat 

orlinmalkja:
AC:

- Translations should be provided as an object in the TIDO config with keys of different languages.
- There should be two (en, de) translation files
- By default TIDO should take the translations from (en) file
- The translations from the translation file and the translation object in TIDO config should be merged. Translations in TIDO config should override their respective keys in translation file
- TIDO translations file should be delivered in (dist) folder of TIDO

### DIFF
--- a/examples/config/4w-local.json
+++ b/examples/config/4w-local.json
@@ -1,6 +1,6 @@
 {
   "showGlobalTree": true,
-  "showNewCollectionButton": true,
+  "showNewCollectionButton": false,
   "theme": {
     "primaryColor": "#1E40AF"
   },
--- a/examples/config/4w-local.json
+++ b/examples/config/4w-local.json
@@ -9,5 +9,14 @@
       "collection": "http://localhost:8181/4w/reproduction/collection.json",
       "manifestIndex": 0
     }
-  ]
+  ],
+  "lang": "en",
+  "translations": {
+    "de": {
+      "sync_panels": "Panels synchronisieren bitte"
+    },
+  "en":  {
+    "sync_panels": "Synchronize panels"
+  }
+  }
 }
--- a/examples/config/4w-local.json
+++ b/examples/config/4w-local.json
@@ -1,6 +1,6 @@
 {
   "showGlobalTree": true,
-  "showNewCollectionButton": false,
+  "showNewCollectionButton": true,
   "theme": {
     "primaryColor": "#1E40AF"
   },
@@ -10,13 +10,5 @@
       "manifestIndex": 0
     }
   ],
-  "lang": "de",
-  "translations": {
-    "de": {
-      "sync_panels": "Panels synchronisieren bitte"
-    },
-  "en":  {
-    "sync_panels": "Synchronize panels"
-  }
-  }
+  "lang": "de"
 }
--- a/examples/config/4w-local.json
+++ b/examples/config/4w-local.json
@@ -10,7 +10,7 @@
       "manifestIndex": 0
     }
   ],
-  "lang": "en",
+  "lang": "de",
   "translations": {
     "de": {
       "sync_panels": "Panels synchronisieren bitte"
--- a/public/translations/de.json
+++ b/public/translations/de.json
@@ -1,6 +1,6 @@
 {
   "new": "Neu",
-  "sync_panels": "Panels Synk",
+  "sync_panels": "Panels synchronisieren",
   "place": "Ort",
   "year": "Jahr",
   "author": "Autor",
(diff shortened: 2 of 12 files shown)


src/utils/config/config.ts
```diff
@@ -71,6 +74,38 @@ function validateTheme(input: any): ValidationResult<AppConfig['theme']> {
...
 }
 
+function validateLang(input: any): ValidationResult<AppConfig['lang']> {
+  const errors: Record<string, string> = {}
+  const result =
+      typeof input === 'string'
+        ? input
+        : (() => {
+          if (input !== undefined)
+            errors['language'] = 'lang must be a string'
```
paulpestov, line 84: this should be `errors['lang']`

src/utils/translations.ts
```diff
@@ -0,0 +1,41 @@
+import i18n from 'i18next'
+import { initReactI18next } from 'react-i18next'
+
+function convertTranslations(translations: Translations) {
+  // translations are converted to expected 3 level format: lang, 'translation', key values translation object
+  // i.e {
+  //      'en': {
+  //          'translation': {'sync_panels': 'Sync Panels'}, {'new': 'New'}
```
paulpestov, line 8: is this correct?

src/App.tsx
```diff
@@ -33,8 +36,16 @@ const App: FC<AppProps> = ({ customConfig }) => {
 
   const collections = useDataStore(state => state.collections)
   const setTreeNodes = useDataStore(state => state.setTreeNodes)
+  const [ready, setReady] = useState(false)
+
 
   useEffect(() => {
+    function initI18nTranslations(translations: Translations) {initI18n(translations)
+      initI18n(translations)
```
paulpestov, line 44: Why can't we just call this outside of useEffect?

src/App.tsx
```diff
@@ -33,8 +36,16 @@ const App: FC<AppProps> = ({ customConfig }) => {
 
   const collections = useDataStore(state => state.collections)
   const setTreeNodes = useDataStore(state => state.setTreeNodes)
+  const [ready, setReady] = useState(false)
```
paulpestov, line 39: Do we need this?

orlinmalkja: After the translations are initialized, then we render App component. To control this I add a state variable `ready`, which is set to `true` after translations are initialized.

src/utils/translations.ts
orlinmalkja: translation is an object, I changed it in my latest commit

orlinmalkja:
@paulpestov I made the necessary changes. Please continue reviewing.

src/utils/config/config.ts
orlinmalkja: yes, changed it

public/translations/de.json
```diff
@@ -0,0 +1,8 @@
+{
+  "new": "Neu",
+  "sync_panels": "Panels Synk",
```
paulpestov, line 3:
```suggestion
  "sync_panels": "Panels synchronisieren",
```

examples/config/4w-local.json
```diff
@@ -9,5 +9,14 @@
       "collection": "http://localhost:8181/4w/reproduction/collection.json",
       "manifestIndex": 0
     }
-  ]
+  ],
+  "lang": "de",
+  "translations": {
```
paulpestov, line 14: no translations overwrite needed in 4w config

tests/cypress/e2e/config.cy.js
```diff
@@ -22,4 +22,12 @@ describe('Config', () => {
   runConfigTest('showNewCollectionButton=false', 'showNewCollectionButton false', () => {
     cy.get('[data-cy="new-collection"]').should('be.visible', false)
   });
+  runConfigTest('', 'translations: No input for `lang` and `translations` object -> read from default `en` file', () => {
```
paulpestov, line 25: this is a default test, you can move it inside the first runConfigTest funtion, so we run all default config tests in one function

orlinmalkja:
@paulpestov I made the necessary changes.

examples/config/4w-local.json
```diff
@@ -1,6 +1,6 @@
 {
   "showGlobalTree": true,
-  "showNewCollectionButton": false,
+  "showNewCollectionButton": true,
```
paulpestov, line 3: why true? this should stay false i think

orlinmalkja: I am not sure if it makes a difference, but changed it.

PR #617, opened 2025-04-15
Collection in collection 

orlinmalkja:


src/contexts/PanelContext.tsx
```diff
@@ -48,7 +47,7 @@ const PanelProvider: FC<PanelProviderProps> = ({ children, panelConfig, index })
     if (!panelId) return
     try {
       setLoading(true)
-      const collection = await getCollection(panelConfig.collection)
+      const { collection } = await getLeafCollection(panelConfig.collection)
```
paulpestov, line 50: Why do we have to use getLeafCollection? Instead, the panel config should already receive the correct collection, as discussed.

paulpestov:
When I run your branch, I see 2 root collections. I think there should be only one?

PR #619, opened 2025-04-16
Default view config

### LINKED ISSUE #604: Add config option to open a specific the view inside a panel
In 4W we want to have the split view as default in order to achieve the Text-Image-Synopsis. So panels should open in that view initially.

AC:
- [] Global config option for setting "split" as default view

paulpestov:
- adds  global "defaultView" config option, so all new opened panels will open with this view
- closes #604

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -79,13 +79,16 @@ declare global {
     primaryColor: string
   }
 
+  type ViewType = 'pip' | 'split' | 'text' | 'image'
+
   interface AppConfig {
     container: string
+    defaultView: ViewType
+    lang: string
     showGlobalTree: boolean
     showNewCollectionButton: boolean
     panels: PanelConfig[]
-    theme: ThemeConfig,
-    lang: string,
+    theme: ThemeConfig
     translations: Translations
   }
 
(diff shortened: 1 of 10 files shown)


src/types.d.ts
```diff
@@ -228,9 +231,9 @@ declare global {
 
   interface PanelConfig {
     collection: string
-    color?: Colors
     manifestIndex?: number
     itemIndex?: number
+    view?: 'pip' | 'split' | 'text' | 'image'
```
orlinmalkja, line 236: i think you should add  as type `ViewType`, since you already defined this new type

src/utils/config/config.ts
```diff
@@ -23,6 +23,19 @@ function validateContainer(input: any): ValidationResult<AppConfig['container']>
   return { result, errors }
 }
 
+function validateDefaultView(input: any): ValidationResult<AppConfig['defaultView']> {
+  const errors: Record<string, string> = {}
+  const result =
+    typeof input === 'string' && ['pip', 'split', 'text', 'images'].includes(input)
```
orlinmalkja, line 29: the last array element should be `image` right ?

paulpestov: done

src/types.d.ts
paulpestov: ah this was a leftover, removed

orlinmalkja (approved):

PR #621, opened 2025-04-17
feat: add placeholder in ImageRenderer, update PanelBody design

paulpestov:
- added placeholder to image renderer
- updated design of panel


orlinmalkja:
I think a few more things need to be improved: 

- Split and text modes should correspond to 2nd and 3rd view respectively.
- Image should extend (become larger) when extending the width of the panel in split mode. Now the image has a fixed width. 
- When there is no image found: When extending the width of the panel, the image icon and the text should keep staying in the center of the panel and not move away from the center or disappear.

orlinmalkja (approved):

PR #623, opened 2025-04-23
feat: update panel content loading, detect more MIME types

paulpestov:


public/translations/de.json
```diff
@@ -13,5 +13,9 @@
   "no_image_available": "Kein Bild verfügbar",
   "could_not_find_image": "Leider konnte kein Bild gefunden werden.",
   "image_preview": "Vorschau des Bildes",
-  "could_not_load_image": "Leider konnte das Bild nicht geladen werden."
+  "could_not_load_image": "Leider konnte das Bild nicht geladen werden.",
+  "empty_panel": "Das ist ein leeres Panel.",
+  "error_occurred": "Ein Fehler ist aufgetreten",
+  "no_content_found": "Keinen passenden Inhalt gefunden.",
+  "unknown_error": "Unbekannter Fehler"
```
orlinmalkja, line 20: should we have an error with this naming ? Shouldn't we always know the type of error which occurred ?

paulpestov: Now it is used in PanelBody as default error message when panelState does not exist which is an extra safety net since panelState can be null.

paulpestov:
fixed

orlinmalkja (approved):

PR #627, opened 2025-04-28
feat: redesign "add new panel" feature, add dialogs for each option

paulpestov:
everything new with add new panel button


src/utils/config/config.ts
```diff
@@ -100,6 +100,19 @@ function validateLang(input: any): ValidationResult<AppConfig['lang']> {
...
 }
 
+function validateTitle(input: any): ValidationResult<AppConfig['title']> {
+  const errors: Record<string, string> = {}
+  const result =
+    typeof input === 'string'
+      ? input
+      : (() => {
+        if (input !== undefined)
+          errors['title'] = 'lang must be a string'
```
orlinmalkja, line 110: the value should be 'Title should be a string'

orlinmalkja (approved):

PR #641, opened 2025-04-29
feat: implement remove panel from actions popover

paulpestov:


orlinmalkja:
There is a bug: when 3 panels are open and I close the leftmost or middle panel, the panel closes. However the close modal is still there on top of another panel.

paulpestov:
fixed

orlinmalkja (approved):

PR #644, opened 2025-04-30
Panel resizing

paulpestov:
AC:

- on first load the loaded panels grow
- without resizing the first: adding new panel will shrink the first panel so both fit in the container
- with resizing the first: the width of that panel will be memorized and the second panel
    -  adapts when there is still space
    - gets the default width when there is no space (=scrollbar)
- having added 4+ panels (so there is no space anymore = scrollbar) the new panel get a default width, scrollbar stays

### DIFF
--- a/src/components/panel/Panel.tsx
+++ b/src/components/panel/Panel.tsx
@@ -15,15 +15,52 @@ const Panel: FC = React.memo(() => {
   const { panelId } = usePanel()
   const scrollPanelIds = useScrollStore(state => state.panelIds)
   const [isScrollPanel, setIsScrollPanel] = useState(false)
+  const [flexValues, setFlexValues] = useState({
+    flexGrow: 1,
+    flexShrink: 1,
+    flexBasis: '0%'
+  })
+
+  const cardRef = useRef(null)
+  const [resizing, setResizing] = useState(false)
+  const [isHoveringEdge, setIsHoveringEdge] = useState(false)
+
+  useEffect(() => {
+    // On mount, we need to decide how set the width. If there is space inside the wrapper,
+    // we can keep default flex settings, as they make the panel grow and use the remaining space.
+    // If there is not enough space, we omit all the growing capability and set a default width
+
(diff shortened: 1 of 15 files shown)


src/components/panel/Panel.tsx
```diff
@@ -15,23 +15,64 @@ const Panel: FC = React.memo(() => {
...
+      .filter(el => el.id !== '' && el.id !== panelId)
+
+    // Return an array of widths. If a panel is in "grow" mode, it has no flexBasis pixel value.
+    // In that case, we set that value to min width as basis for total width calculation.
+    // The reason is that those panels can also shrink.
+    const otherPanelWidths = otherPanelEls.map(el => {
+      return (el.style.flexBasis.includes('px')) ? parseInt(el.style.flexBasis.replace('px', '')) : MIN_PANEL_WIDTH
+    })
+
+    const panelsTotalWidth = otherPanelWidths.reduce((a, b) => a + b, 0)
```
orlinmalkja, line 47: i think the calculated value is otherPanelsTotalWidth, right ?

src/components/panel/Panel.tsx
```diff
@@ -15,23 +15,64 @@ const Panel: FC = React.memo(() => {
...
+
+    // We are adding here an additional DEFAULT_PANEL_WIDTH for the ghost panel that is the last child in the wrapper.
+    const isFitting = (wrapperWidth - (panelsTotalWidth + DEFAULT_PANEL_WIDTH)) >= 0
```
orlinmalkja, line 50: Maybe the condition should be `>=DEFAULT_PANEL_WIDTH` ? Since the current Panel would fit if the remaining width is at least equal as DEFAULT_PANEL_WIDTH.

orlinmalkja (approved):

src/components/panel/Panel.tsx
paulpestov: It would also fit if its less that default if the old panels haven't been resized so one or more of them can shrink and make space. Let's try this for now.

PR #647, opened 2025-05-02
 improve the design of metadata heading on hover

### LINKED ISSUE #634: Fix: When hovering on a heading in metadata, then it should not underline, since it bring c
When hovering on a heading in metadata it should not underline, since it might bring confusion to the users as they are links and could lead to i.e another page. Instead when hover the following interaction should happen:

- Heading should not underline
- Should be like in https://headlessui.com/v1/vue/disclosure

orlinmalkja:
Closes #634

### DIFF
--- a/src/components/metadata/CollectionMetadata.tsx
+++ b/src/components/metadata/CollectionMetadata.tsx
@@ -21,9 +21,8 @@ const CollectionMetadata: FC = () => {
 
   return <Accordion type="single" collapsible>
     <AccordionItem value="item-1" >
-      <AccordionTrigger className="t-py-1 t-relative [&>svg]:t-absolute [&>svg]:t-right-3">{ t('collection') }</AccordionTrigger>
-      <AccordionContent><div className="t-ml-2">{ metadata.map((meta, i) => <MetadataItem item={meta} key={i} />)}</div>
-      </AccordionContent>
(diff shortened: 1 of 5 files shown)


src/components/metadata/CollectionMetadata.tsx
```diff
@@ -21,8 +21,8 @@ const CollectionMetadata: FC = () => {
 
   return <Accordion type="single" collapsible>
     <AccordionItem value="item-1" >
-      <AccordionTrigger className="t-py-1 t-relative [&>svg]:t-absolute [&>svg]:t-right-3">{ t('collection') }</AccordionTrigger>
-      <AccordionContent><div className="t-ml-2">{ metadata.map((meta, i) => <MetadataItem item={meta} key={i} />)}</div>
+      <AccordionTrigger className="t-py-[4px] t-pl-3 t-pr-1 t-mr-[6px] t-rounded-md t-bg-gray-100 hover:t-no-underline hover:t-bg-gray-200 t-mb-1 t-text-gray-700 t-font-bold">{ t('collection') }</AccordionTrigger>
```
paulpestov, line 24: Please follow [DRY](https://de.wikipedia.org/wiki/Don%E2%80%99t_repeat_yourself) and implement it directly in the accordion comps

PR #666, opened 2025-05-20
Add easier navigation inside a panel

orlinmalkja:


src/components/panel/ItemLabel.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <Button
+            variant={showItemModal ? 'secondary' : 'ghost'}
+            className="font-semibold text-gray-600"
+            onClick={() =>  setShowItemModal(!showItemModal)}
+            data-cy="item-label">
+            { getItemLabel() }
+          </Button>
+        </PopoverTrigger>
+        {showItemModal && <PopoverContent side="bottom" align="start" sideOffset={8} className="flex flex-col space-y-2 max-w-[200px] w-fit max-h-[450px] h-fit pr-0 pl-2 py-2">
+          <div className="text-gray-600 ml-1">Please select an item to open</div>
```
paulpestov, line 101: translate

src/components/panel/ItemLabel.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <div className="text-wrap">
+            <div className="flex flex-col space-y-1 max-h-[350px] overflow-y-auto">
+              {labels.length > 0 && labels.map((label, i) => <Button
+                variant="ghost"
+                style={{ color: 'Page '+ label === getItemLabel() ? primaryColor : '' }}
```
paulpestov, line 106: please do a better check here, "Page" may change, also use a Tailwind class for primaryColor instead of the variable.

src/components/panel/ManifestLabel.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <PopoverTrigger asChild>
+          <Button
+            variant={showManifestModal ? 'secondary' : 'ghost'}
+            className="relative font-semibold text-gray-600"
+            onClick={() =>  setShowManifestModal(!showManifestModal)}
+            data-cy="manifest-label">
+            { selectedLabel }
+            { selectedManifest && <div
+              className="absolute mb-4 right-1 flex items-center justify-center w-4 h-4 border-[1px] border-gray-400 text-yellow-300 text-md font-bold rounded-full">
+                !
```
paulpestov, line 60: Use an icon here

src/components/panel/ManifestLabel.tsx
```diff
@@ -0,0 +1,84 @@
...
+            </div> }
+          </Button>
+        </PopoverTrigger>
+        {showManifestModal && <PopoverContent side="bottom" align="start" sideOffset={8} className="flex flex-col space-y-2 max-w-[350px] w-fit max-h-[450px] pr-0 h-fit pl-2 py-2">
+          <div className="text-gray-600 ml-1">Please select a manifest to open</div>
```
paulpestov, line 65: translate

src/components/panel/PanelTitle.tsx
```diff
@@ -1,26 +1,32 @@
...
   }
 
+
   return (
     <>
       <div className="flex items-center">
         { (!panelState || !panelState.item) && <Skeleton className="w-[100px] h-6" />  }
-        { panelState && panelState.item && <span className="font-semibold">{ getManifestLabel() }</span>}
-        <span className="w-[1px] h-[80%] bg-muted mx-2 grow-0 shrink-0"></span>
+        { panelState && panelState.item  && <ManifestLabel showManifestModal={showManifestModal} setShowManifestModal={setShowManifestModal} updateSelectedManifest={updateSelectedManifest}  setShowItemModal={setShowItemModal} selectedManifest={selectedManifest.current} />}
```
paulpestov, line 26: too many props, why are they needed? can't the logic be moved into label comps?

src/components/panel/ItemLabel.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const collection = useDataStore().collections[panelState.collectionId]?.collection
+  const manifest = panelState.manifest
+  const updatePanel = usePanelStore(state => state.updatePanel)
+
+  const [labels, setLabels] = useState([])
+  const externallyOpened = useRef(false)
+
+
+  useEffect(() => {
+    async function getLabels() {
```
paulpestov, line 29: why the function?

src/components/panel/ManifestLabel.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <div className="text-wrap">
+            <div className="flex flex-col space-y-1 max-h-[350px] overflow-y-auto">
+              {labels.map((label, i) => <Button
+                variant="ghost"
+                key={i}
+                style={{ color: label === selectedLabel ? primaryColor : '' }}
```
paulpestov, line 71: same here as above, omit primaryColor variable

src/components/panel/ItemLabel.tsx
```diff
@@ -0,0 +1,119 @@
...
+          </Button>
+        </PopoverTrigger>
+        {showItemModal && <PopoverContent side="bottom" align="start" sideOffset={8} className="flex flex-col space-y-2 max-w-[200px] w-fit max-h-[450px] h-fit pr-0 pl-2 py-2">
+          <div className="text-gray-600 ml-1">Please select an item to open</div>
+          <div className="text-wrap">
+            <div className="flex flex-col space-y-1 max-h-[350px] overflow-y-auto">
+              {labels.length > 0 && labels.map((label, i) => <Button
+                variant="ghost"
+                style={{ color: 'Page '+ label === getItemLabel() ? primaryColor : '' }}
+                key={i} className="text-wrap h-fit min-h-8 overflow-hidden justify-start pl-2 mr-2 py-1"
```
paulpestov, line 107: use combination of label + i for key

orlinmalkja:
@paulpestov please continue reviewing

src/components/panel/PanelTitle.tsx
```diff
@@ -1,26 +1,29 @@
...
     <>
       <div className="flex items-center">
         { (!panelState || !panelState.item) && <Skeleton className="w-[100px] h-6" />  }
-        { panelState && panelState.item && <span className="font-semibold">{ getManifestLabel() }</span>}
-        <span className="w-[1px] h-[80%] bg-muted mx-2 grow-0 shrink-0"></span>
+        { panelState && panelState.item  && <ManifestLabel updateSelectedManifest={updateSelectedManifest}  setShowItemModal={setShowItemModal} selectedManifest={selectedManifest.current} />}
+        <span className="w-[1px] h-[80%] bg-gray-400 mx-2 grow-0 shrink-0"></span>
         { (!panelState || !panelState.item) && <Skeleton className="w-[40px] h-6" />  }
-        { panelState && panelState.item && <span className="text-muted-foreground" data-cy="item-label">{ getItemLabel() }</span>}
+        { panelState && panelState.item && <ItemLabel selectedManifest={selectedManifest.current} updateSelectedManifest={updateSelectedManifest}  showItemModal={showItemModal} setShowItemModal={setShowItemModal} />}
```
paulpestov, line 26: why are the modals controlled externally and not inside of the label components?

src/components/panel/ManifestLabel.tsx
```diff
@@ -0,0 +1,82 @@
...
+          </Button>
+        </PopoverTrigger>
+        {showModal && <PopoverContent side="bottom" align="start" sideOffset={8} className="flex flex-col space-y-2 max-w-[350px] w-fit max-h-[450px] pr-0 h-fit pl-2 py-2">
+          <div className="text-gray-600 ml-1">{ t('please_select_a_manifest_to_open') }</div>
+          <div className="text-wrap">
+            <div className="flex flex-col space-y-1 max-h-[350px] overflow-y-auto">
+              {labels.map((label, i) => <Button
+                variant="ghost"
+                key={label + '_'+i}
+                className={`h-fit min-h-8 justify-start overflow-hidden pl-2 mr-2 ${panelState.manifest.label === label ? 'text-sky-600': ''} `}
```
paulpestov, line 70: why sky-600? please use primary or so

src/components/panel/ManifestLabel.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <>
+      <Popover open={showModal} onOpenChange={handleOpenChange}>
+        <PopoverTrigger asChild>
+          <Button
+            variant={showModal ? 'secondary' : 'ghost'}
+            className="relative font-semibold text-gray-600"
+            onClick={() =>  setShowModal(!showModal)}
+            data-cy="manifest-label">
+            { selectedLabel }
+            { selectedManifest && <CircleAlert className="absolute mb-4 right-1 text-yellow-300" /> }
```
paulpestov, line 60: this still looks odd, the border is gray, just check online how such tags are styled.

src/components/panel/ManifestLabel.tsx
```diff
@@ -0,0 +1,82 @@
...
+          </Button>
+        </PopoverTrigger>
+        {showModal && <PopoverContent side="bottom" align="start" sideOffset={8} className="flex flex-col space-y-2 max-w-[350px] w-fit max-h-[450px] pr-0 h-fit pl-2 py-2">
+          <div className="text-gray-600 ml-1">{ t('please_select_a_manifest_to_open') }</div>
```
paulpestov, line 64: don't use gray, but rather muted or secondary foregrounds

paulpestov:
please merge the dark mode here and check for design refactors

src/components/panel/PanelTitle.tsx
orlinmalkja: Because Manifest modal can trigger the display of the ItemModal (an external open). Therefore I place setShowItemModal() in a parent component  - PanelTitle.

orlinmalkja:
> please merge the dark mode here and check for design refactors

the dark mode is already in next branch, what do you exactly mean ?

src/components/panel/ManifestLabel.tsx
paulpestov: icon is good, but use a slightly darker color, yellow-500

paulpestov:
please check the hover state on active items, primary color should not disappear

src/components/panel/ManifestLabel.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { apiRequest } from '@/utils/api.ts'
+
+
+interface ManifestLabelProps {
+  selectedManifest: Manifest | null,
+  updateSelectedManifest: (newManifest: Manifest | null) => void
+  setShowItemModal: (show: boolean) => void,
+}
+
+const ManifestLabel: FC<ManifestLabelProps> = ({ selectedManifest, updateSelectedManifest, setShowItemModal }) => {
```
paulpestov, line 18: setShowItemModal shouldn't be part of that components logic, label comps should only open/close and select, any logic beyond should be handled in the parent comp

orlinmalkja:
@paulpestov I think I addressed all your points !

src/components/panel/PanelTitle.tsx
```diff
@@ -1,26 +1,33 @@
...
     <>
       <div className="flex items-center">
         { (!panelState || !panelState.item) && <Skeleton className="w-[100px] h-6" />  }
-        { panelState && panelState.item && <span className="font-semibold">{ getManifestLabel() }</span>}
-        <span className="w-[1px] h-[80%] bg-muted mx-2 grow-0 shrink-0"></span>
+        { panelState && panelState.item  && <ManifestLabel onManifestSelect={onManifestSelect}  selectedManifest={selectedManifest.current} />}
+        <span className="w-[1px] h-[80%] bg-gray-400 mx-2 grow-0 shrink-0"></span>
         { (!panelState || !panelState.item) && <Skeleton className="w-[40px] h-6" />  }
-        { panelState && panelState.item && <span className="text-muted-foreground" data-cy="item-label">{ getItemLabel() }</span>}
+        { panelState && panelState.item && <ItemLabel selectedManifest={selectedManifest.current} updateSelectedManifest={updateSelectedManifest}  showItemModal={showItemModal} setShowItemModal={setShowItemModal} />}
```
paulpestov, line 30: the API in the ManifestLabel is good, please stick to it on ItemLabel also

paulpestov:
why are the arrow moving after selecting the manifest? this should not happen

src/components/panel/ItemLabel.tsx
```diff
@@ -0,0 +1,119 @@
...
+      item: newItem,
+      config: { ...panelState.config, manifestIndex: manifestIndex, itemIndex: itemIndex }
+    })
+
+    setShowItemModal(false)
+    onItemSelect()
+  }
+
+  function getItemLabel() {
+    return 'Page ' + (panelState?.item?.n ?? 'unknown')
```
paulpestov, line 76: Just for quick improvement, so we don't need a task later: Could you use a translation here for "Page" please? `t('item')`

src/components/panel/ManifestLabel.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <>
+      <Popover open={showModal} onOpenChange={handleOpenChange}>
+        <PopoverTrigger asChild>
+          <Button
+            variant={showModal ? 'secondary' : 'ghost'}
+            onClick={() =>  setShowModal(!showModal)}
+            title={selectedLabel}
+            data-cy="manifest-label"
+            className="px-2 py-1 h-7 w-[150px]"
```
paulpestov, line 55: please don't use heights (and widths) for text items

src/components/panel/ItemLabel.tsx
```diff
@@ -0,0 +1,119 @@
...
+              </div>
+            </Button>
+          </div>
+        </PopoverTrigger>
+        {showItemModal && <PopoverContent side="bottom" align="start" sideOffset={8} className="flex flex-col space-y-2 max-w-[200px] w-fit max-h-[450px] h-fit pr-0 pl-0 py-2">
+          <div className="text-wrap text-muted-foreground ml-1 mb-0 pl-1">{ t('please_select_an_item_to_open') }</div>
+          <div className="flex flex-col space-y-1 max-h-[350px] overflow-y-auto pb-1 pt-1 pl-1">
+            {labels.length > 0 && labels.map((label, i) => <Button
+              variant="ghost"
+              key={label + '_'+ i} className={`h-8 overflow-hidden justify-start mr-2 py-1 pl-2 ${panelState.item.n === label ? 'text-primary hover:text-primary': ''}`}
```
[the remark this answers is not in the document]
paulpestov: True, but then a button is not a good choice here. Also for a11y reasons.

paulpestov:
Looking closer into this I really think you should use the Dropdown components for the lists. As shadcn states a Popover is a " rich content in a portal" while a Dropdown Menu is "a menu to the user — such as a set of actions or functions". The lists a literally menus that users can select actions from. If you look into the HTML of a dropdown menu comp then you see that they set all required a11y properties such as "role=menuitem" which is the correct way to do it. In our current case we just have a list of buttons. Imagine a screen reader trying to parse our app, he will repeatately mention "Button with text XY", "Button with text YZ" 100 times which is not what we want. 

Switching to dropdowns shouldn't be a big deal since you already have a good API on the label comps. You can just replace the inner stuff while keeping the same logic as it was before.

src/components/panel/ItemLabel.tsx
orlinmalkja: Yes I added it

paulpestov:
Thanks, the dropdowns look good I only commented a small issue! But unfortunately the widths still don't work, for example on large screens: 

image

Any width setting is the wrong path here because the texts are dynamic as mentioned before.

PR #677, opened 2025-06-13
Produce an after build tido.plain css file

orlinmalkja:
In this PR I aim to produce a tido css file without '@layer', instead containing only plain css. Postcss is a recommended way to do this, since it is more reliable than using 'regex' expression. 

I add a postcss command in vite config, which runs after build (closeBundle) and converts tido.min.css to plain css file `tido.min.plain.css`


orlinmalkja:
@paulpestov Feel free to test it in an example project not using Tailwind 4.

vite.embed.config.js
```diff
@@ -16,7 +18,19 @@ export default defineConfig(({ mode}) => {
       react(),
       tailwindcss(),
       ...(env.VITE_ENV === 'production' ? [removeAttrs(['data-cy'])] : []),
-      injectConfig(projectName)
+      injectConfig(projectName),
+      {
```
paulpestov, line 22: Please move the plugin to a separate file like "injectConfig" and just call it in the vite config. The reason is that we need that plugin also in vite.config.js file which builds the standalone bundle. As we deliver both, I didn't find any better way to handle this double-build as having two vite configs. I know there are duplicates in both configs, this is still pending to resolve.

PR #683, opened 2025-06-18
Add css variable conversion to rgb values to use tido colors independent of project's css framework

orlinmalkja:
The approach of removing 'root' and 'host' from `tido.min.css` file didn't work.

In this PR I aim to solve the scope color issue in TIDO by converting Tailwind CSS color variables to rgb values.

We are scoping tido colors under .tido{} class. The project which implements TIDO does not know about CSS variables of TIDO, since they are isolated. Therefore the TIDO colors should have a _universal color value_ - rgb value,  independent of project's css framework.


paulpestov:
Hm I don't think is a good idea, I don't want to manage Tailwind colors. Also RGB has nothing to do with our issue, it's just a different color format, oklch is still fine.

PR #690, opened 2025-06-20
Add make view configurable

### LINKED ISSUE #686: Make views configurable
AC: 

- Add new config key "views"
- The value should be an array that allows string values `swap`, `split`, `text`, `image`
- Default config should be all four
- The existance of a view string in the config should show/hide text view button on top right of a panel
- Optional: the order of the array is also represented in the UI

- Very optional: it should be guaranteed that no other source can manipulate the display of the text view buttons and so no display of the text should be allowd when that view is disabled by config. Example: project has configured 2 views: split and image, bookmarking url says it should open "swap" -> Result: swap view shouldn't be possible, swap button shouldn't exist

orlinmalkja:
Closes #686

### DIFF
--- a/src/components/panel/select-view/Views.tsx
+++ b/src/components/panel/select-view/Views.tsx
@@ -1,37 +1,44 @@
 import { FC, useState } from 'react'
 import { useTranslation } from 'react-i18next'
+import { useUIStore } from '@/store/UIStore.tsx'
+import { useConfigStore } from '@/store/ConfigStore.tsx'
 
 import { AlignCenter, Columns2, Image, PictureInPicture2 } from 'lucide-react'
-import { ViewType } from '@/types'
+import { ViewButtonData, ViewType } from '@/types'
 import { Button } from '@/components/ui/button.tsx'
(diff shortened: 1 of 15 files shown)


paulpestov:
I think we should completely omit the viewIndex and use the string name everywhere instead.

src/components/panel/select-view/Views.tsx
```diff
@@ -13,9 +15,10 @@ interface SelectViewButtonsProps {
 const Views: FC<SelectViewButtonsProps> = ({ updateSelectedButton }) => {
   const [selectedView, setSelectedView] = useState(useUIStore.getState().defaultView)
   const { t } = useTranslation()
+  const views = useConfigStore(state => state.config.views)
 
-  const buttonsData = {
-    pip: {
+  const defaultButtonsData = {
```
paulpestov, line 20: For `buttonData` I would drop the object structure and turn it into an array, if we now want to rely on the order.

paulpestov (approved):

PR #694, opened 2025-06-24
Add smart display of views

### LINKED ISSUE #689: Smart display of views
In general TIDO should follow a stretegy of being smart about the displayed content. This means that we should offer a good solution in various situations automatically. 

One feature in that field could be to decide which view is the most suitable for a initial display based on our TextAPI responses. 

AC:
- Disable (not hide) view buttons on the top right of a panel that are not suitable according to our information
- Diable swap view when no image is present in the item response
- Disable split view when no image is present in the item response
- Disable image view view when no image is present in the item response
- Load the first suitable view 
- There should be a priority hierarchy of which view to load, ideas:
   - "defaultView" config option should have highest prio but when it's not suitable  choose a different one
   - order of the "views" config option should come next

orlinmalkja:
Closes #689

### DIFF
--- a/src/components/panel/TextViewsToggle.tsx
+++ b/src/components/panel/TextViewsToggle.tsx
@@ -51,8 +51,9 @@ const TextViewsToggle: FC = () => {
     updatePanel(panelState.id, { view: newView })
   }
 
-  function isEnabled(buttonView: ViewType) {
-    return !(!panelState.imageExists && ['image', 'swap', 'split'].includes(buttonView))
+  function isDisabled(buttonView: ViewType) {
+    return (!panelState.imageExists && ['image', 'swap', 'split'].includes(buttonView))
+      || !panelState.contentTypes?.length
   }
 
 
@@ -71,7 +72,7 @@ const TextViewsToggle: FC = () => {
                     onClick={(e) => handleTextViewClick(e, button.view)}
                     variant={panelState.view === button.view ? 'secondary' : 'ghost'}
                     size="icon"
-                    disabled={!panelState.contentTypes?.length || !isEnabled(button.view)}
+                    disabled={isDisabled(button.view)}
                     data-selected={panelState.view === button.view}
                     data-cy={button.view}
                   >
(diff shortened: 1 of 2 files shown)


src/types.d.ts
```diff
@@ -146,7 +146,8 @@ declare global {
     view: ViewType
     imageUrl?: string
     activeTargetIndex: number
-    config: PanelConfig
+    config: PanelConfig,
+    existsImage: boolean
```
paulpestov, line 150: Rename to imageExists

src/components/panel/TextViewsToggle.tsx
```diff
@@ -65,7 +67,7 @@ const TextViewsToggle: FC = () => {
                     onClick={(e) => handleTextViewClick(e, button.view)}
                     variant={panelState.view === button.view ? 'secondary' : 'ghost'}
                     size="icon"
-                    disabled={!panelState.contentTypes?.length}
+                    disabled={!panelState.contentTypes?.length || (!panelState.existsImage && ['image', 'swap', 'split'].includes(button.view))}
```
paulpestov, line 70: Move this to a function `isEnabled`, reverse the condition, we should always stay positive :)

src/components/panel/TextViewsToggle.tsx
```diff
@@ -65,7 +71,7 @@ const TextViewsToggle: FC = () => {
                     onClick={(e) => handleTextViewClick(e, button.view)}
                     variant={panelState.view === button.view ? 'secondary' : 'ghost'}
                     size="icon"
-                    disabled={!panelState.contentTypes?.length}
+                    disabled={!panelState.contentTypes?.length || !isEnabled(button.view)}
```
paulpestov, line 74: This should also into isEnabled, so the function becomes a central place to determine whether a button is enabled or not

src/components/panel/TextViewsToggle.tsx
```diff
@@ -50,6 +51,11 @@ const TextViewsToggle: FC = () => {
     updatePanel(panelState.id, { view: newView })
   }
 
+  function isEnabled(buttonView: ViewType) {
+    return !(!panelState.imageExists && ['image', 'swap', 'split'].includes(buttonView))
```
paulpestov, line 55: Hm this is less helpful than I thought, maybe change it back to your original condition and name the function ìsDisabled`

paulpestov (approved):

PR #695, opened 2025-06-25
Improve Dialog component 

orlinmalkja:


src/components/ui/dialog.tsx
```diff
@@ -47,28 +43,42 @@ const DialogOverlay = ({
...
+const DialogContent = ({
+  className,
+  children,
+  customOverlay,
+  showClose = true,
+  ...props
+}: DialogContentProps) => {
+  return (
+    <>
+      {customOverlay ? customOverlay : <DialogOverlay />}
```
paulpestov, line 63: I think the DialogPortal needs to be dealt with still. The dialog in `AddPanel` is buggy, maybe because it is appended to the panel wrapper. In general, all dialogs should be the highest visible item, this is the reason why we hide everything behind the DialogOverlay. Users are "forced" to clear it first before continuing. To ensure this we should by default append the dialogs to `.tido` and in special cases receive a custom container.

orlinmalkja:
@paulpestov please continue reviewing!

paulpestov (approved):

PR #735, opened 2025-08-21
Fix images not loading

### LINKED ISSUE #730: Images are not loading
When opening the TIDO-Viewer in https://test.vierwachen.d.sub.uni-goettingen.de/edition the images are not loaded. According to Michelle there is no request from TIDO but the entry in the TextAPI is available and the images can be reached via URL.

orlinmalkja:
Closes #730


paulpestov:
Change requests to fulfill the current task: 

- Please use "resetBoundary" instead of key for item id changes
- Please use the same error component design as TextViewError

Problems that remain:
- Timing issue at PanelContext, loading state should go into the PanelState and thus being read from in all consumer comps
- Local loading state in ImageRenderer is a redundant state, this should be rather that central loading state from above

orlinmalkja:
@paulpestov please continue reviewing

PR #741, opened 2025-08-26
fix: check the response of annotation collection

### LINKED ISSUE #736: Handling text that has no annotations
In Neologie, there are collections/manifests that do not have any annotations (E.g. [BdN III Griesbach ](https://bdnportal.d.sub.uni-goettingen.de/edition-viewer/griesbach/v5) "Textkritische Ansicht"). 

At the moment, TIDO5 does not load any content from this manifest because it seems to expect annotations.

orlinmalkja:
Closes #736

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,6 +43,7 @@
         "postcss-cli": "^11.0.1",
         "react-error-boundary": "^6.0.0",
         "react-i18next": "^15.4.1",
+        "sonner": "^2.0.7",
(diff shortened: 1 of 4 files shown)


orlinmalkja:
In my last commit I do the following:
-  catch all errors in toast
- to provide a title and description to toast, I add a new Error class, namely CustomError which contains 'name' and 'message' attributes

orlinmalkja:
@paulpestov I think I made the necessary changes!

PR #749, opened 2025-09-03
Read and update annotations in PanelState after updating other Panel states' attributes

### LINKED ISSUE #746: Text view does not update on annotation collection error
**Describe the bug**
When we navigate to an item that doesn't have annotation collection (ex: Textkritische), we see the error message but the Text view does not get's update and clicking on Next Item button doesn't update text view as well.

**To Reproduce**
1. Go to Manifest Textkritische
2. Click on first item which doesn't have annotations collection

**Expected behavior**
Text view should be update and should still be able to navigate through items even without annotations.

**Additional context**
May be disable Annotations Icon as well from Panel Modes ?

orlinmalkja:
- read and update annotations in PanelState after updating other panel states' attributes

an error in annotation collection should not prevent us from updating current Panel State with item, manifest, content types etc

Closes #746


paulpestov:
Haha good idea, but still dangerous since we don't control the exact timings of the actual updates of the visible content. The root problem is that everything is happening within one single function call. There should be a better strategy how to get all the different data, for example get the data when a specific part of the panel is about to be mounted. 

Good for now.

PR #754, opened 2025-09-05
feat: add variants display in annotation items, witness filter in sid…

paulpestov:
- witness chips appear in annotation items
- tooltip over chips shows full title
- witness filter appears in sidebar header when variants are avaiable
- annotation items + highlighting is toggled on/off when changing the witness filter
- witness filter is persisted across all content inside panel

### DIFF
--- a/src/components/panel/annotations/AnnotationContent.tsx
+++ b/src/components/panel/annotations/AnnotationContent.tsx
@@ -1,0 +1,57 @@
+import React, { FC } from 'react'
+import { parseStyleString } from '@/utils/html-to-react.ts'
+
+interface Props {
+  value: string
+}
+
+const convertNodeToReact = (node, key) => {
+  if (node.nodeType === Node.TEXT_NODE) {
+    return node.textContent
+  }
+
+  if (node.nodeType !== Node.ELEMENT_NODE) return null
+
+  const Tag = node.tagName.toLowerCase()
+  const children = Array.from(node.childNodes).map((child, i) =>
+    convertNodeToReact(child, `${key}-${i}`)
+  )
+
+  const props = {}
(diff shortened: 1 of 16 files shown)


orlinmalkja:
It looks really good! I will add a few comments.

src/components/panel/annotations/WitnessFilter.tsx
```diff
@@ -0,0 +1,59 @@
...
+const WitnessFilter: FC = () => {
+  const { witnesses, selectedWitnesses, setSelectedWitnesses, usePanelTranslation, matchedAnnotationsMap, setMatchedAnnotationsMap } = usePanel()
+  const { t } = usePanelTranslation()
+
+  function toggleSelectedWitness(witness: WitnessWithColor, checked: boolean) {
+    const selected = [ ...selectedWitnesses ]
+    if (checked) selected.push(witness)
+    else {
+      const index = selected.findIndex(w => w.idno === witness.idno)
+      if (index > -1) selected.splice(index, 1)
```
orlinmalkja, line 21: Is this check necessary ? What we click in witnesses filter, it belongs to witnesses right ?

paulpestov: yes it's just one of those syntacital "null checks", the context of your code might change but your function should still work...

src/components/panel/annotations/AnnotationContent.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { FC } from 'react'
```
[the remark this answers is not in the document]
paulpestov: Yes, we need to use central functions for generating html string to react. I would like to address that in a separate PR.

PR #757, opened 2025-09-09
Add cross ref improvement issue #752

### LINKED ISSUE #752: Cross ref improvement
Michelle suggested to add a config option to open the referenced element: either in this Panel or in a new Panel.

Another idea is NOT to use a config option, but always to give the opportunity to user on element click by showing a dropdown. If we deal with an internal link the dropdown items should be in following order: 'Open in this Panel' and 'Open in new Panel'.
In the case of an external link then the dropdown should contain the options in order 'Open in new Panel' and 'Open in this Panel'.

orlinmalkja:
Closes #752


src/utils/other-functions.ts
paulpestov, line 1: Personally, I don't like common "util" files because it will end up as a trash bin. If you use that function once in your component, but it doesn't fit into the comp body then put it outside the comp but in the same file. If you need it in several comps then put it into "dom.ts" util file or something

src/components/panel/CrossRef/CrossRefSameItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+      {link}
+    </DropdownMenuTrigger>
+    <DropdownMenuContent align="center" className="p-2">
+      <DropdownMenuLabel>
+        {t('scroll_to')}
+      </DropdownMenuLabel>
+      <CrossRefTitle contentType={node.getAttribute('data-ref-content-type')} manifestLabel={manifestLabel} itemLabel={itemLabel} />
+      <DropdownMenuSeparator></DropdownMenuSeparator>
+      <DropdownMenuItem asChild className="mt-2 cursor-pointer">
+        <span onClick={() => navigateToSameItem(node, contentTypes, 'scroll-to')}>{t('scroll_to')}</span>
```
paulpestov, line 75: Why custom onClick? Isn't there a native onSelect prop on DropdownMenuItem?

src/components/panel/CrossRef/CrossRefDifferentItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const { t } = usePanelTranslation()
+
+  const collectionId = node.getAttribute('data-ref-collection')
+  const manifestIndex = useRef(null)
+  const manifest = useRef(null)
+  const itemIndex = useRef(null)
+  const item = useRef(null)
+
+  const itemLabel = useRef<string>(null)
+  const manifestLabel = useRef<string>(null)
```
paulpestov, line 35: Why are they refs and not states?

public/translations/de.json
```diff
@@ -60,6 +60,7 @@
     "please_select_panel_mode": "Bitte wählen Sie einen Modus",
     "show_image": "Zum Bild",
     "show_text": "Zum Text",
+    "scroll_to": "Scrollen Sie zu",
```
paulpestov, line 63:
```suggestion
    "reference": "Verweis",
```

src/components/panel/CrossRef/CrossRefDifferentItem.tsx
orlinmalkja: When CrossRefDifferentItem is mounted then the newManifestLabels and newItemLabels are updated. I use only one boolean to rerender the component and pass the updated props to CrossRefTitle component.

Why is it necessary to use states ?

PR #771, opened 2025-09-17
Check for witness value in witness chip

### LINKED ISSUE #770: Bug: Not found witness causes error
**Describe the bug**
A not found witness causes an error

**To Reproduce**
1. Navigate to London Brit. 7200, Item 114v
2. Switch to transcription content
3. Open annotations panel
4. Navigate to next item

 **Actual behavior**
TIDO crashes

**Expected behavior**
To navigate in new item AND load correctly the witnesses

**Additional context**

orlinmalkja:
Closes #770

### DIFF
--- a/src/components/panel/annotations/WitnessChip.tsx
+++ b/src/components/panel/annotations/WitnessChip.tsx
@@ -7,19 +7,21 @@ interface Props {
 }
 
 const WitnessChip: FC<Props> = React.memo(({ idno }) => {
-  const { witnesses } = usePanel()
+  const { witnesses, usePanelTranslation } = usePanel()
+
+  const { t } = usePanelTranslation()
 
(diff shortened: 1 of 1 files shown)


src/components/panel/annotations/WitnessChip.tsx
```diff
@@ -10,16 +10,16 @@ const WitnessChip: FC<Props> = React.memo(({ idno }) => {
   const { witnesses } = usePanel()
 
   function getWitnessLabel(idno: string) {
-    return witnesses.find(w => w.idno === idno).idnoAlt ?? 'none'
+    return witnesses.find(w => w.idno === idno)?.idnoAlt ?? 'none'
```
paulpestov, line 13: i think we have a translation for "unknown", maybe it fits here

PR #790, opened 2025-09-30
Display an error node on api request failure of global tree nodes

### LINKED ISSUE #753: Tree Node Api Error
When a tree node produces an api request error i.e on manifest or collection url, then instead of a child element, a custom element should be rendered which contains
- Error message
- Retry button which gives the opportunity to try to reload the children nodes

orlinmalkja:
Closes #753

### DIFF
--- a/src/components/tree/ErrorNode.tsx
+++ b/src/components/tree/ErrorNode.tsx
@@ -1,0 +1,23 @@
+import { FC } from 'react'
+import { CircleX, RotateCcw } from 'lucide-react'
+import { Button } from '@/components/ui/button.tsx'
+import { useTranslation } from 'react-i18next'
+
+interface Props {
+  onRetry: () => Promise<void>
+}
+
+const ErrorNode: FC<Props> = ({ onRetry }) => {
+  const { t } = useTranslation()
+
+
+  return <div className="flex flex-col p-2 ml-12 rounded-xl">
+    <span className="mb-2 flex">
+      <CircleX className="text-red-500 mr-2" size="20" />
+      <span data-cy="error-node" className="text-red-800/60 dark:text-red-50/50">{t('could_not_load_data')}</span>
+    </span>
+    <Button variant="destructive" className="w-[65%] active:scale-95 transition duration-150 ease-in-out" onClick={onRetry}><RotateCcw /> { t('retry') }</Button>
+  </div>
+}
+
+export default ErrorNode
--- a/src/components/tree/TreeNode.tsx
+++ b/src/components/tree/TreeNode.tsx
@@ -4,32 +4,50 @@ import { useTranslation } from 'react-i18next'
 import { ChevronRight, Folder, LibraryBig, File, FolderOpen } from 'lucide-react'
 
 import EmptyNode from '@/components/tree/EmptyNode.tsx'
+import ErrorNode from '@/components/tree/ErrorNode.tsx'
 
 
 interface TreeNodeProps {
   node: TreeNode,
 }
 
+
 const TreeNode: FC<TreeNodeProps> = ({ node }) => {
   const { onSelect, getChildren, selectedNodeId, setSelectedNodeId } = useTree()
   const { t } = useTranslation()
   const [isExpanded, setIsExpanded] = useState(node.expanded)
   const [showEmptyNode, setShowEmptyNode] = useState(false)
+  const [showErrorNode, setShowErrorNode] = useState(false)
+  const [children, setChildren] = useState(node.children)
+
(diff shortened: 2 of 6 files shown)


orlinmalkja:
@paulpestov I made your suggested changes!

src/components/tree/TreeNode.tsx
```diff
@@ -4,41 +4,67 @@ import { useTranslation } from 'react-i18next'
...
+      } catch (error) {
+        console.error(error)
+        if (!showErrorNode) setShowErrorNode(true)
+        return
+      }
+    }
 
-    if (node.children.length > 0) {
+    if (!node.leaf && isExpanded) {
       toggleExpand()
```
paulpestov, line 45: Why is toggleExpand called again? Maybe toggle is not the right function, should probably split it up into "expand/collapse" functions

src/components/tree/TreeNode.tsx
```diff
@@ -4,41 +4,67 @@ import { useTranslation } from 'react-i18next'
...
 }
 
+
 const TreeNode: FC<TreeNodeProps> = ({ node }) => {
   const { onSelect, getChildren, selectedNodeId, setSelectedNodeId } = useTree()
   const { t } = useTranslation()
   const [isExpanded, setIsExpanded] = useState(node.expanded)
   const [showEmptyNode, setShowEmptyNode] = useState(false)
+  const [showErrorNode, setShowErrorNode] = useState(false)
+  const [children, setChildren] = useState(node.children)
```
paulpestov, line 21: Is this still necessary? It wasn't before, was it? node.children would have been updated and it worked.

src/components/tree/TreeNode.tsx
```diff
@@ -4,41 +4,67 @@ import { useTranslation } from 'react-i18next'
...
-  const toggleExpand = () => {
-    setIsExpanded(!isExpanded)
+
+  async function onErrorRetry () {
+    try {
+      // setChildren -> rerender the component
+      setChildren([...await getChildren(node)])
+      setShowErrorNode(false)
+    } catch(error) {
+      console.error(error)
```
paulpestov, line 66: Shouldn't it set the error state again to true?

src/components/tree/TreeNode.tsx
```diff
@@ -4,41 +4,67 @@ import { useTranslation } from 'react-i18next'
...
-    if (!node.expanded && !node.leaf) node.children = [...await getChildren(node)]
+    if (!isExpanded && !node.leaf) {
+      try {
+        toggleExpand()
+        node.children = [...await getChildren(node)]
+        setChildren(node.children)
+        if (node.children.length > 0) return
+      } catch (error) {
+        console.error(error)
+        if (!showErrorNode) setShowErrorNode(true)
```
paulpestov, line 39: Why the if?

src/components/tree/TreeNode.tsx
paulpestov: Ok, so then call it just once on not leafs.

src/components/tree/TreeNode.tsx
paulpestov: So line 64 should come before calling getChildren to reset the error state before trying to reload the children. Then if it fails again it should go into error state again. This is for example useful if we want to display a loading spinner in between.

src/components/tree/TreeNode.tsx
orlinmalkja: yes

src/components/tree/TreeNode.tsx
orlinmalkja: The code is the previous implementation of handleNodeClick

src/components/tree/ErrorNode.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+const ErrorNode: FC<Props> = ({ onRetry }) => {
+  const { t } = useTranslation()
+
+
+  return <div className="flex flex-col p-2 ml-12 rounded-xl">
+    <span className="mb-2 flex">
+      <CircleX className="text-red-500 mr-2" size="20" />
+      <span data-cy="error-node" className="text-red-800/60 dark:text-red-50/50">Failed to fetch data</span>
```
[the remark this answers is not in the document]
orlinmalkja: added translation

src/components/tree/TreeNode.tsx
orlinmalkja: yes updated setShowError accordingly

paulpestov (approved):

PR #794, opened 2025-10-06
feat: add first implementation of bookmarking

paulpestov:


orlinmalkja:
please check linting problem

src/contexts/ConfigContext.tsx
```diff
@@ -0,0 +1,65 @@
+import { createContext, useContext, use } from 'react'
+import { TidoConfig } from '@/types'
+import { mergeAndValidateConfig } from '@/utils/config/config.ts'
+import { promiseWithCache } from '@/utils/promise-for-suspense.ts'
```
orlinmalkja, line 4: please check this util file, I think you should change the util filename to 'promise-cache'. I get error while building the project, since this file name is not existing

orlinmalkja:
I know that this is a draft PR - also work in progress, but I thought to still add the following error here
image

PR #799, opened 2025-10-07
fix: elevation hierarchy for backgrounds and hover effects

### LINKED ISSUE #716: Dark mode: No hover effect in different places inside Panel
While in dark mode inside the Panel there is no hover effect i.e lighter background color in the following parts.

- local tree
- manifest labels in dropdown
- item labels in dropdown

A suggestion of the possible hover effect can be as shown in screenshot.

Image

paulpestov:
Aligning elevation colors like it was intended by shadcn. 

Background -> Muted -> Accent

Examples: 

- Global Tree - background color is "background", so hover color is "muted"
- Collection Title Popover - background color is (same as) "muted", so hover color is "accent"

Also made annotation type filters as "primary" when selected.

closes #716

### DIFF
--- a/src/components/tree/LocalTree.tsx
+++ b/src/components/tree/LocalTree.tsx
@@ -18,7 +18,6 @@ const LocalTree: FC<Props> = ( { collectionId, onSelect } ) => {
   const updatePanel = usePanelStore(state => state.updatePanel)
   const [treeNodes, setTreeNodes] = useState([])
 
-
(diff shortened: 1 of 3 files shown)


orlinmalkja:
Is it a good design practice regarding buttons to have more than 1 primary button in one page ? We already have 'Add new Panel'.

paulpestov:
> could you please apply hover effect on 'local tree' in _dark mode_ as well ?

fixed

PR #812, opened 2025-10-16
feat: add general panel error component, catch panel init errors

### LINKED ISSUE #628: Bug: Incorrect error message when creating a new panel via a new 'parent' collection
When creating a new panel via a new 'parent' collection an error arises: "Cannot read properties of undefined (reading 'map')".

Instead it should show as error message: "Please provide a leaf collection which contains a sequence of manifests"

paulpestov:
Error from cannot be reproduced but the main topic was still not handled yet. The problem is that we don't have any general panel error handling, only through toast. This PR introduces an ErrorBoundary for that. As discussed earlier, the setup is still not ideal since we lack a Panel Shell which would take care of width resizing. That's why we have to set it's own width on the PanelError comp.

Try to load https://bdnportal.d.sub.uni-goettingen.de/api/textapi/noesselt/collection.json as panel config or any other wrong URLs 

closes #628

### DIFF
--- a/src/components/panel/PanelError.tsx
+++ b/src/components/panel/PanelError.tsx
@@ -2,7 +2,7 @@ import { FC, useEffect, useState } from 'react'
 import { CircleX } from 'lucide-react'
 import { FallbackProps } from 'react-error-boundary'
 import { CustomError } from '@/contexts/PanelContext.tsx'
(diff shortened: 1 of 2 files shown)


orlinmalkja:
When I provide a random string in collection, then I receive the following error. The error in such case should be better caught/ better reported.
image

paulpestov:
Done with fixes

Issue #734, opened 2025-08-21
Extended Split Mode


paulpestov:
EUPT Req: 2 texts side-by-side from same item, synced, each text has own annotations

The idea is to have a combination of config options that replicate TIDO v4 behaviour. Projects can pre-configure their desired panel setup. The functionality of adding new panels dynamically should be disabled everywhere. There should be a config option to turn on the "sync navigation mode", so all UI parts behave accordigly, for example the global tree should not ask where to open a new item but rather reload all panels with that item. The panel configs themselves should be very flexible on defining the visible content. These config options should be re-applied when the content switches, for example upon selecting in the global tree.

paulpestov:
Update from refinement: 

We will try to implement the Extended Split-View because adding an additional hierarchy layer (groups) may distract users in recognizing related texts. If going the "split" approach, the multiple texts still remain in the ame panel and thus have a common panel header with navigation and other options. 

Following the requierements could be identified for the first implenmentation:

- A new panel mode will be introduced that has X amount of split parts
- Image and text views can be loaded into the split parts
- Multiple text/images can be displayed
- Annotations for all visible views appear in the sidebar
- It is possible to preconfigure the split parts

PR #832, opened 2025-11-10
feat: add resizable panes for every panel content element, remove sel…

### LINKED ISSUE #714: Resize Annotation Panel
Issue written from Yana:

As a user, I want to be able to resize the unfolded annotation panel horizontally to the left, so that long annotations are easier to read without excessive scrolling.

AC
- The unfolded annotation panel can be resized horizontally towards the left by dragging its left edge or a visible resize handle.
- When hovering over the resizable edge/handle, the cursor changes to a horizontal resize icon (↔).
- The cursor only changes when resizing is possible.
- A minimum and maximum width is set.
- Resizing happens in real-time as the user drags the panel edge.
- Text and annotation content reflow dynamically within the panel during resizing.

paulpestov:
New features:
- New library "allotment" is now used to create resizable panel contents
- There are 2 resizable panes per default: main and sidebar
- The library calculates automatically the sizes of the panes by setting width pixel values dependant on the parent container.
- The panel-resizer is still alive but with very reduced usage. It just resizes the outer panel width now.
- We use a "hack" to keep the main pane unchanged when we open the sidebar - opening the sidebar means enlarging the parent container which leads to recalculation of the main pane by the lib. That's why we remember the current width of the main pane and set it as minSize and maxSize (recommended by the lib) to keep it fixed. After the opening of the sidebar has finished we change minSize/maxSize to the normal settings again.
- With this panes setup the text and annotation views become separate in scrolling. There comes the SidebarScroller into play. He syncs the the scrolling between the sidebar and then declared text.
- Since the whole container situation has changed with the lib, there is a new structure in rendering the panel now. A new comp PanelShell is now handling the resizing and displaying the outer border. PanelContext is not showing error boundary but passing the error down which is a break the the "boundary pattern" but in this case it a decision towards simplicity. The reason is that the PanelShell also needs variables from PanelContext. 
- Removed SelectPanelMode popup feature. Yes, we could make it configurable, but I think the popup in general was not a good idea. Maybe we can rework this feature by opening the already existing dropdown if ever needed again. 
- Removed the old sync panel feature since there has to be refactored a lot anyway in the future. 

Future options:

- The main pane is now also divided into resizable panes. The content for each pane is now determined by an array. So it becomes more dynamic in terms of defining in how many parts the split mode should split. This is a preparation for #734. 

closes #714

### DIFF
--- a/src/components/panel/Swapper.tsx
+++ b/src/components/panel/Swapper.tsx
@@ -1,4 +1,4 @@
-import { FC } from 'react'
+import { FC, ReactNode } from 'react'
 import { Image, Text } from 'lucide-react'
 import { Button } from '@/components/ui/button.tsx'
 import { usePanel } from '@/contexts/PanelContext.tsx'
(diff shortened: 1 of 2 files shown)


orlinmalkja:
Another point is to remove  'shade' effect in 'Witness filter', right ?
Besides that, the PR looks good!

Issue #824, opened 2025-10-30
Use resource ids instead if indexes


paulpestov:
Currently we are using indexes for items and panels at several places. Above all it is the panel config that relies heavily on those. It would be better and more flexible to use the respective resource URLs directly. This way we stabilize our API and make it more consistent if taking the bookmarking as one "channel" and the config as the other channel of our API.

PR #848, opened 2025-11-19
No indexes in config

paulpestov:
- all configs have to use resource ids now
- new possible scenarios like only item was provided or manifest -> item do not fit together

closing #824


src/components/panel/CrossRef/CrossRefDifferentItem.tsx
```diff
@@ -79,32 +79,26 @@ const CrossRefDifferentItem: FC<Props> = ({ node }) => {
...
-
       updatePanel({
         config: {
           collection: collectionId,
-          manifestIndex,
-          itemIndex,
-          contentIndex: newContentIndex
+          manifest: manifestId,
+          item: itemId,
+          activeContentType: contentType
```
orlinmalkja: the key should be 'contentType'

PR #849, opened 2025-11-20
Selection marks global tree

### LINKED ISSUE #615: Selection marks in global tree
New feature: When users work with the global tree they will open and close it multiple times. Upon selecting something from the tree selection popup the active state vanishes. Therefore the global tree should "remember" which items are currently visible in which panels. A small visible UI element should indicate that a respective items is already loaded in a panel. 

AC:

- [ ] Upon selecting an item a small (green?) circle is display at the right of the tree node and all parent nodes
- [ ] Hovering over the circle displays a tooltip (title attribute) which the panel index or description
- [ ] Clicking on the circle makes the respective panels scroll intro view

orlinmalkja:
Closes #615

### DIFF
--- a/src/components/tree/OpenedIcon.tsx
+++ b/src/components/tree/OpenedIcon.tsx
@@ -2,6 +2,7 @@ import { FC } from 'react'
 
 import { Circle } from 'lucide-react'
 import { useTranslation } from 'react-i18next'
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip.tsx'
 
 interface Props {
   panelsNumbers: number[],
@@ -30,9 +31,17 @@ const OpenedIcon: FC<Props> = ({ panelsNumbers, nodeType }) => {
   }
 
 
-  return <span title={getTooltip(nodeType, panelsNumbers)}>
-    <Circle className="text-green-500 fill-green-500 hover:scale-130 transition-all duration-300" size={10} fill="true" onClick={(e) => scrollToPanel(e)}  />
-  </span>
+  return <><TooltipProvider delayDuration={600}>
+    <Tooltip>
+      <TooltipTrigger asChild>
+        <Circle className="text-green-500 fill-green-500 hover:scale-130 transition-all duration-300" size={10} fill="true" onClick={(e) => scrollToPanel(e)}  />
+      </TooltipTrigger>
+      <TooltipContent>
+        <span className="leading-none">{ getTooltip(nodeType, panelsNumbers) }</span>
+      </TooltipContent>
+    </Tooltip>
+  </TooltipProvider>
+  </>
 }
 
 export default OpenedIcon


paulpestov:
Some design fixes should be made:
- the green dot should be aligned vertically across all tree nodes
- it should sit vertically in the middle of a tree node itself
- there should be enough padding to the right edge of the tree node, so in hover state the dot is not on the edge (probably the absolute positioning is not a good option, but rather define clear areas within a tree node like "collape icon area", "label area", "actions/options area on the right")
- there should be no position change (jumping) 
- the dot could be a little bit smaller
- small hover effect?

paulpestov:
Is it possible to use our shadcn tooltip component instead of the title attribute?

paulpestov:
Ideas for additions later:

- Instead of the current tooltip, a popover could open upon hover/click with a preview UI representing the current panels (similar to the global tree modal), there uses can click on each "panel button" to jump to it
- A loading indicator could be renderer as soon as the user has made a selection in the tree, the indicator could sit the at the same position as the green but look like a "round skeleton", after the panel has finished loading the skeleton would be come green

src/components/tree/TreeNode.tsx
```diff
@@ -38,6 +45,42 @@ const TreeNode: FC<TreeNodeProps> = ({ node }) => {
...
+        }
+      })
+      setPanelsNumbersOpened(newPanelsNumbersOpened)
+    }
+
+
+    async function updatePanelsNumbersForCollection(node: TreeNode) {
+      const newPanelsNumbersOpened = []
+      const rootCollection = await apiRequest<Collection>(node.id)
+      const childCollectionIds = await getRootChildrenCollectionsIds(rootCollection)
```
paulpestov: Please see comment above, is this necessary?

src/components/tree/TreeNode.tsx
```diff
@@ -86,20 +130,26 @@ const TreeNode: FC<TreeNodeProps> = ({ node }) => {
...
         <div className={`shrink-0 mt-1 mx-2 ${node.leaf ? 'ml-5': ''}`}>
           { node.type === 'collection' && <LibraryBig size={18} />}
           { node.type === 'manifest' && !isExpanded && <Folder size={18} />}
           { node.type === 'manifest' && isExpanded && <FolderOpen size={18} />}
           { node.type === 'item' && <File size={18} />}
         </div>
-        <span data-cy="node-label">{node.label}</span>
+        <span data-cy="node-label" className="w-[80%] truncate">{node.label}</span>
+        {panelsNumbersOpened.length > 0 && <div data-cy="tree-node-actions" className="grow relative flex items-center h-[100%]">
+          <div className="absolute right-0 top-1/2 -translate-y-1/2">
```
paulpestov: Does this have to be absolute positioned now? Please have a deeper look on flexbox layouts before position something absolutely.

PR #859, opened 2025-11-28
test: add annotationsMode config tests

jfrer:
Adds tests for the config option annotationsMode


jfrer:
Draft because of #861

PR #869, opened 2025-12-09
Content type in cross ref

### LINKED ISSUE #850: Content type shown in cross reference info
When a project has implemented cross references, clicking on a link creates an info box about where this link leads and in which panel the text should be opened:

Image

Here, the content type ("edition") is also stated. This is not necessary since the content type is an internal, custom MIME type that might have a value that doesn't necessarily fit. Also these are often stated in English while the text might be in another language as is the case above.

IMHO the content type can be completely left out.

paulpestov:
closes #850

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,27 +62,6 @@
         "react-dom": "^19.1.0"
       }
     },
-    "node_modules/@asamuzakjp/css-color": {
-      "version": "3.2.0",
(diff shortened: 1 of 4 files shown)


orlinmalkja:
@paulpestov I think the issue is different. I understand it as following: The content type should not appear at all in Cross Ref Popover.

PR #896, opened 2026-01-20
test: add config tests for language translation keys and for panel modes

jfrer:
Adds these testcases: 

- Configure custom common translation
- Configure custom common translation for custom translation key
- Configure custom collection translation for custom translation key
- Configure multiple panel modes
- Configure only one panel mode
- Configure a default panel mode and a panel specific panel mode

### DIFF
--- a/tests/cypress/e2e/config.cy.js
+++ b/tests/cypress/e2e/config.cy.js
@@ -1,3 +1,4 @@
+import { Panel } from '../support/panel-helpers'
 function runConfigTest(param, name, callback, only=false) {
   const test = () => {
     cy.visit('/e2e.html?' + param);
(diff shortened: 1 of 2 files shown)


tests/cypress/e2e/config.cy.js
```diff
@@ -376,5 +379,40 @@ describe('Config', () => {
...
+  });
+  runConfigTest('lang=de&translations.de.reproduction.accurate=genau&panels[0].collection=http://localhost:8181/4w/reproduction/collection.json',
+    'Should apply custom translation "genau" for custom translation key "accurate" and collection key "reproduction"', () => {
+      cy.get('[data-cy="content-type"]').should('have.text', 'genau')
+  });
+  runConfigTest('panelModes[]=text&panelModes[]=split',
+    'Should have panel mode "text" preselected, with "split" as the only other selectable panel mode', () => {
+      getPanelModeOption('text')
+        .should('have.attr', 'data-selected', 'true')
+        .parents('[data-cy="panel-mode-menu"]').focus().type('{esc}')
```
paulpestov: Why is this necessary? Also another point is possibly to avoid parent/child selections. I noticed this during fixing the tests for the panel header PR. It's tidious to find the reason for a failing test when selections follow a strict path of nestings. Since it will be a common thing to select a certain panel mode, maybe it would be a good idea to refactor this area a bit and use it in other tests, too.

Issue #887, opened 2025-12-19
New annotation structure for variants


mweidling:
After some internal discussion we decided on a new annotation structure / model for variants.
This is the **old** structure:

```json
{
  "@context" : "http://www.w3.org/ns/anno.jsonld",
  "partOf" : {
    "label" : "Ahiqar annotations for textgrid:3r678: Paris, Bibliothèque Nationale de France, ms. syr. 434",
    "id" : "http://ahiqar.uni-goettingen.de/ns/annotations/annotationCollection/3r678"
  },
  "next" : "https://api.ahiqar.sub.uni-goettingen.de/annotations/ahiqar/syriac/3r678/187r/latest/annotationPage.json",
  "type" : "AnnotationPage",
  "prev" : null,
  "id" : "http://ahiqar.uni-goettingen.de/ns/annotations/annotationPage/syriac/3r678-186v",
  "items" : [{
    "body" : {
      "x-content-type" : "Variant",
      "type" : "TextualBody",
      "format" : "text/plain",
      "value" : {
        "witnesses" : [ "t_162" ],
        "entry" : "qwerty"
      }
    },
    "target" : [ {
      "selector" : {
        "type" : "CssSelector",
        "value" : "#abc123"
      }
    }, {
      "language" : "syc",
      "id" : "http://ahiqar.uni-goettingen.de/ns/annotations/3r676/-to-",
      "format" : "text/xml"
    }, {
      "selector" : {
        "type" : "TextQuoteSelector",
        "exact" : ""
      },
      "source" : "http://example.org/page1"
    } ]
}],
  "refs" : [ {
    "idnoAlt" : "J",
    "title" : "Jerusalem, SMMJ 162",
    "tguri" : "3r673",
    "idno" : "t_162"
  }]
}
```

The **new** structure will look like this:

```json
{
  "@context" : "http://www.w3.org/ns/anno.jsonld",
  "partOf" : {
    "label" : "Ahiqar annotations for textgrid:3r678: Paris, Bibliothèque Nationale de France, ms. syr. 434",
    "id" : "http://ahiqar.uni-goettingen.de/ns/annotations/annotationCollection/3r678",
    "refs" : [ { // refs moves to "partOf" since the page does not allow it in the way we had it before. "refs" has to be added to the annotation collection as well, but this shouldn't bother TIDO
      "idnoAlt" : "J",
      "title" : "Jerusalem, SMMJ 162",
      "tguri" : "3r673",
      "idno" : "t_162"
[...]

Issue #706, opened 2025-07-22
Implement annotation filters


paulpestov:
Annotations can have different types. Sometimes users don't want to see every type in the sidebar at the same time. Therefore we need a filter to toggle the visiblity of annotations with selected types.

<img width="860" height="307" alt="Image" src="https://github.com/user-attachments/assets/cb49a713-15fd-47d3-8958-6cd830f72ba0" />

AC:

- A filter bar is displayed in the sidebar header of a panel
- The types are read from `body['x-content-type']` of a annotation
- All types are selected per default
- Clicking on a filter button toggles the visiblity of respective annotations in the sidebar body
- By toggling the visilibty, the yMap has to be updated, so all remaining annotations sit at the right spots
- When toggling on certian types, additional filters may appear below the filter bar
- Type "Variant" should show a witnesses select below
- Each annotation item should render a small type ribbon on the top left
- Optional/To be tested: The ribbon should only appear on select, yMap has to update because the selected annotation grows larger

PR #902, opened 2026-01-26
New annotation filters

paulpestov:
This PR implements a first version of the annotation filters as discussed in #706.

Unfortunately we cannot test the configuration in different projects as we don't have updated data yet. There was a change in the TextAPI #887. So we have to wait until we there a valid backend for this. But until then we can test basic annotation types. See `ahiqar-local.json` for a config example.

What this PR adds:

- A new config key "annotations" that gives the ability to define flexible setups for annotations. It contains two kinds of configs "types" and "filters". Types is a map where the key equals the annotation type (x-content-type) and the value contains a config object for the respective type. Filters is a hierarchical definition of the filters config. 
- There are two elements now in the annotation header, filters button and list/aligned toggle.
- Filters are opened in a popover.
- Filters are not fully generic. We differ between root level and all other levels. Under "filters.rootSelectionRule" users can toggle the behaviour o nthe root level to be "single" or "multiple". Single will render a tab bar on top which switches the tree of deeper filters. Multiple will just render a tree.
- Please note that in the config we deal with one recursive object of a filter tree but the selection rule decides how to render it. This is due to keep the config users focused on their hierarchical structure and not bother about rendering details. 
- While in single mode we remember the selected tree in each root category. So switching between root nodes won't reset the trees. 
- It is possible to configure multiple annotation types for a single filter node. This has reasons, so users can combine filtering by multiple annotation types at the same time, which reapplies the feature for TIDO4 (panel tabs). 
- We do **not** imply any configuration. Since config options can be quite arbitrary I chose to validate and assume as little as possible because it is a hell to debug. Therefore we take the config almost as-is. This means  config users have to define their selected states with caution. We also don't have indeterminate checkbox state yet. 
- We still work with the AnnotationTypeDict which is the main interface for annotation types that should be visible. Now it is just updated by having always an array as value. This is due to Variants. Witnesses will appear as array items in the dict. 
- Variant witnesses have to configured explicitely, so there is control over display of each witness.

### DIFF
--- a/src/components/panel/annotations/filters/AnnotationFilters.tsx
+++ b/src/components/panel/annotations/filters/AnnotationFilters.tsx
@@ -1,4 +1,4 @@
-import { FC } from 'react'
+import { FC, useEffect } from 'react'
 import { usePanel } from '@/contexts/PanelContext.tsx'
 import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover.tsx'
 import { Button } from '@/components/ui/button.tsx'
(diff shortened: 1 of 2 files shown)


orlinmalkja:
When specifiying a certain number of filters in root level i.e 7 of them then the last one i.e 'Dictionary' can't be seen well. Maybe introduce kind of horizontal scrollbar or so ? Please see the picture below. 
image

orlinmalkja:
As a general remark: I think introducing 'types' and 'filters' separately makes the annotations filter concept more complex and more difficult to understand. I think a better option would be to stick with one key - which should design how the annotation types including filters should look like. This would lead also to a more general representation of Annotation Types Filtering.

paulpestov:
> As a general remark: I think introducing 'types' and 'filters' separately makes the annotations filter concept more complex and more difficult to understand. I think a better option would be to stick with one key - which should design how the annotation types including filters should look like. This would lead also to a more general representation of Annotation Types Filtering.

I think the advantage of having a separate "types" key is that it is flat, as the annotationPage is in the TextAPI. So you can define a config for types without a hierarchy. Filters are meant to be nested, so there I would lkike to focus on the structure itself.

orlinmalkja (approved):
**Please notice:** The update of annotations filter content on item navigation is still not fixed. The list of annotations doesn't always represent the filters selection above. 
I can take care of this after this PR then

PR #914, opened 2026-02-03
Handle long annotations

### LINKED ISSUE #837: Handle long annotation contents
Annotation can have very long contents. There should be a mechamism to extend/collapse an annotation.

AC
- By default display some specific height for every annotation
- For longer annotations not selected: make overflow hidden of text and add 'View more' button. Height of annotation remains same
- Selecting a longer annotation or clicking 'View more' shows fully the content of annotation
- Selecting a new annotation or deselecting active long annotation restores the annotation content to default

orlinmalkja:
Closes #837

### DIFF
--- a/src/components/panel/annotations/Annotation.tsx
+++ b/src/components/panel/annotations/Annotation.tsx
@@ -136,8 +136,8 @@ const Annotation: FC<Props> = React.memo(({ data, top, onExpand, onCollapse }) =
         { type === 'Variant' && <VariantContent body={data.body} /> }
         { type !== 'Variant' && <AnnotationContent body={data.body} /> }
       </div>
-      { isLong && !isExpanded && <Button className="w-fit h-2 mt-4 px-0" variant="text" onClick={(e) => handleViewMore(e)} >{t('view_more')}</Button> }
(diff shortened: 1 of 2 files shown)


paulpestov:
I think the general idea works now. But the implementation is a nightmare for the browser. On my machine at least, the transitions do not run smoothly, especially on collapsing. In BDN we have lots of annotations, so this issue will be very present there. Unfortunately, we don't have reliable test data to this moment. 

So, I could live also without transitions of the height, would it improve the performance? 
Moreover there is still the issue that the expansion itself is not performed well. You see how the text shows up immediately at full possible height but the annotation border is trying to catch up. This is even more noticable when the annotation is very long, so the animation has "travel" a longer distance. Usually you want to set overflow:hidden to the "about-to-expand" container which creates the effect of actual expanding. But as I said, I think it's fine just snap it directly to the full height.

Also some details that I don't understand: 

- Why is `${isLong && !isExpanded ? 'h-18 overflow-y-hidden' : 'h-fit'}` still present in the Annotation classes when we are setting pixel values in the callbacks?
- I see you update `desiredY` in onAnnotationExpand directly by applying the height delta which makes totally sense and is probably more efficient than calling `trackTopChange`. Why are you calling `trackTopChange` in onAnnotationCollapse instead of doing the same thing backwards? Wouldn't this be also more efficient?

paulpestov (approved):

PR #920, opened 2026-02-10
New panel modes

### LINKED ISSUE #734: Extended Split Mode

paulpestov:
closes #734 

- removed old panel modes
- introduced new config "panelViews", please see default-config and types 
- new "Views" dropdown in the panel header
- annotations from all split panes are now displayed together
- filters are applied on all split panes

### DIFF
--- a/src/contexts/TextViewContext.tsx
+++ b/src/contexts/TextViewContext.tsx
@@ -55,6 +55,7 @@ export const TextViewProvider = ({
 
   // Once the higher up map is updated, update the local map
   useEffect(() => {
+    if (!Object.hasOwn(matchedAnnotationsMaps, activeContentUrl.current)) return
     setMatchedAnnotationsMap(matchedAnnotationsMaps[activeContentUrl.current])
   }, [matchedAnnotationsMaps, setMatchedAnnotationsMap])
 


paulpestov:
> I obtain following bug when navigating many times to next item or when refreshing multiple times the page

Can you provide repro steps? I cannot reproduce.

orlinmalkja:
> > I obtain following bug when navigating many times to next item or when refreshing multiple times the page
> 
> Can you provide repro steps? I cannot reproduce.

Yes sure
1. Run app with `ahiqar.json` file and following `panelViews`: 
2.  "panelViews": [ 
        { "label": "Diplomatic", "view": "text", 
        "contentTypes": ["transcription", "translation"] 
      }, 
     { 
      "label": "Text", "view": "text", "contentTypes": ["transcription"] 
    }, 
    { 
     "label": "Image", "view": "image"
     }, 
    { "label": "Text", "view": "text", "contentTypes": ["transcription", "translation"] } 
   ],
3. Either Keep refreshing the page and I obtain following error 

<br />

image

<br />

4. or keep navigating to next item and I obtain the above error

orlinmalkja:
> > Bug 1:
> > Bug 2:
> > Bug 3:
> 
> should be fixed now

Yes these Bugs are fixed. Thanks! However the 2 newly added Bugs remain in this PR

orlinmalkja:
I get the following Bug with `ahiqar-local` data
1. run `ahiqar-local`
2. We already have opened British museum 7209, 182b 
3. Click next item

Actual behaviour 
image

PR #929, opened 2026-02-16
Config annotations mode

### LINKED ISSUE #916: Add config option to disable list mode in annotations
AC: 

- add config option "annotations.mode" that allows values of either "list" or "aligned"
- if the option is not provided at all, we keep everything as-is (mode toggle is displayed)
- if option is provided, the mode toggle is not displayed and annotations are rendered in the respective mode automatically

orlinmalkja:
I tried to extend more the config for annotations mode by introducing in total **two keys as optional** . With this we get a more generic configuration of `AnnotationsMode`.  

An additional reason for introducing two keys is in order to actually differentiate between the cases when the user wants the annotations toggle with an initial value or only one mode without the toggle.

- `singleMode` - we show the annotations only in this specified mode AND remove the AnnotationsModeToggle
- `defaultMode` - we show the toggle with the initial selected value as specified with this value

Things to consider
- If no `singleMode` is specified then annotationsConfig obtains `defaultMode` which is specified from TIDO internally
- If nothing is specified then the same happens as above.

Closes #916

### DIFF
--- a/src/utils/config/config.ts
+++ b/src/utils/config/config.ts
@@ -1,7 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-const DEFAULT_ANNOTATIONS_MODE = 'aligned'
-
 import { defaultConfig } from '@/utils/config/default-config.ts'
 
 import enTranslations from '../../../public/translations/en.json'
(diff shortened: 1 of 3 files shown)


src/utils/config/config.ts
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+const DEFAULT_ANNOTATIONS_MODE = 'aligned'
```
paulpestov: This should probably come from default config. We do set values from there already somewhere in the validation.

PR #944, opened 2026-02-27
Ready event

### LINKED ISSUE #826: Ready event
The app should emit a "ready" event as soon as all asynchronous operations have been finished. This is useful in various secenarios for the future.

orlinmalkja:
Closes #826

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -324,11 +324,6 @@ declare global {
     x: number,
     y: number
   }
-
-  interface TidoPublicApi {
-    onReady?: () => void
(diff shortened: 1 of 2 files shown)


src/types.d.ts
```diff
@@ -325,6 +325,10 @@ declare global {
     y: number
   }
 
+  interface TidoPublicApi {
```
paulpestov: Call this `TidoInstance` since it will be the type of the instance variable. Also, since you are introducing this interface, what about encodeState and decodeState?

PR #950, opened 2026-03-04
refactor: use native resizing of sidebar instead of nested Allotment panes

paulpestov:
- Removed nested Allotment setup. Previously the sidebar and main content would sit in 2 Allotment panes while the main content was an Allotment container itself containing the content views. The sidebar was mounted upon state change which caused friction while resizing the panel width. This was solved by a "hack" where we would set a fixed width to main content's pane during resizing and unset it when resizing was finished. 
- Now the sidebar is an absolutely position element again and the resizing between main content and sidebar is done by the panel-resizer.


orlinmalkja:
Could you please add a description in this PR, like which are the exact changes ?

PR #957, opened 2026-03-12
Annotations in annotations

### LINKED ISSUE #874: Annotations in annotations

orlinmalkja:
This is a first implementation of `Annotations in Annotations` feat. 
Please notice the following
1. I considered only a single selection of target in this PR. I didn't consider a second selection, which can select the next annotation in multi-annotations target or deselect it.
2. I did not aim to solve the expanding of annotations inside another annotation in Align Mode. The changes I include here should improve a bit the current solution and pave the way to the next PR on actually using this feature in Align Mode.

Closes #874

### DIFF
--- a/src/utils/annotations.ts
+++ b/src/utils/annotations.ts
@@ -69,9 +69,79 @@ function computeNewSelectedAnnotationIndex(targetEntry: MergedAnnotationEntry, p
   return newSelectedAnnotationIndex
 }
 
+function getNestedAnnotations(annotation: Annotation, itemAnnotations: Annotation[]) {
+  if (itemAnnotations.length === 0) return []
+  return itemAnnotations.filter((annot)  => annot.target[0].source === annotation.id)
+}
+
+function findTargetsInsideAnnotation(annotationId: string, itemAnnotations: Annotation[]) {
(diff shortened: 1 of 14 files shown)


paulpestov:
Set padding bottom

image

src/utils/annotations.ts
```diff
@@ -69,9 +71,111 @@ function computeNewSelectedAnnotationIndex(targetEntry: MergedAnnotationEntry, p
...
+    selectors.push(selector)
+  })
+
+  return {
+    elements,
+    selectors
+  }
+}
+
+function getFlippedNestedMatchedAnnotationsMap(nestedMatchedAnnotationsMap) {
```
paulpestov: use typing, unnecessary comment?

orlinmalkja:
@paulpestov I made your requested changes!

paulpestov:
> I fixed the bug: "When selecting a parent annotation with expanded nested annotations (child annotations), the highlighting in the text disappears."

I'm not sure if this is fixed: 

- Run dev example
- Click on "in possession of" in the right text

New highlighting appears in the left text. How?

Issue #871, opened 2025-12-11
Tooltip Annotations


paulpestov:
There should be another way to display annotations which is through tooltips. When users hover or click on a target within a text a tooltip with the annotation content should appear. This is useful for "dictionary" type of annotations or annotations that appear in lots amouts where they would be polluting the annotation panel. 

Detailed feature description TBD.

paulpestov:
First Implementation will contain AC from MEKONG sprint: 

- There is a config option to set one or multiple annotation types as "to be shown as tooltips".
- On click of the target a tooltip appears.
- Tooltip annotations are not reflected in the filters (later maybe add as feature "tooltips on/off")
- Only one tooltip is open at a time.
- Works for 4W (links to FWB) as HTML link.
- It is visually clear that it is a functioning and clickable link

PR #984, opened 2026-04-08
feat: add tooltip annotations

paulpestov:
implements first part of #871, see comment for AC


orlinmalkja:
Bug: 
1. we open the tooltip
2. select outside the tooltip
3. the target keeps the hover style -> instead on popover close when we are not hovering on the target, the hover style should be removed

orlinmalkja:
Additional Bug: 
1. Hover on a target
2. Mouse leave target
3. The hover style remains, instead it should be removed

orlinmalkja:
In AnnotationSidebar for '4 Wachen' it says that there are not annotations related to this text. This is not correct, since there are annotations, but they are not shown in sidebar.

paulpestov:
> Additional Bug:
> 
> 1. Hover on a target
> 2. Mouse leave target
> 3. The hover style remains, instead it should be removed

Cannot reproduce

paulpestov:
> In AnnotationSidebar for '4 Wachen' it says that there are not annotations related to this text. This is not correct, since there are annotations, but they are not shown in sidebar.

Yes, we might deal with this later. Look at AC 3 at the issue. It suggests to introduce a "tooltips on/off" option. This would intergrate tooltip annotations as part of annotations. Right now, we want to completely ignore those and just display them on spot.

orlinmalkja (approved):

PR #985, opened 2026-04-09
Generic text rendering - generic component which includes main text functionalities of TextRenderer

orlinmalkja:
Closes #976 
We add a component GenericTextRenderer which renders HTML text and CrossRefs Nodes. In addition it takes care of following text-annotation functionalities

- Highlights targets of an html string given its annotations
- Hovers bidirectionally targets and annotations
- New for child annotations: Multi-annotation target selection: Handles the mechanism when click at a target
- New: now we use only the related annotations for the text i.e for Annotation text we use only its child annotations to highlight targets and create MatchedMap

Any HTML string which has a certain set of annotations can use the above functionalities by only using GenericTextRenderer.


src/components/panel/annotations/AlignAnnotationsList.tsx
```diff
@@ -130,14 +130,18 @@ const AlignAnnotationsList: FC = () => {
...
+        }
         return {
           target,
           el,
           desiredY: target.offsetTop,
           annotation
         }
       })
-      setElements(_elements)
+      setElements(_elements.filter(Boolean))
```
paulpestov, line 144: wtf :D

paulpestov (approved):

PR #991, opened 2026-04-15
feat: add target-based synced scrolling in text when scrolling the sidebar

paulpestov:
closes #972 

- open bdn
- scroll in sidebar
- targets will align to 35% of the text's height

there is 1 focused annotation, it is identified when it's body intersects 35% of sidebar's height, then the text's target is being aligned

things to improve later: 
- make the target "drive" together with the annotation in the sidebar when no other annotation intersects the line
- scroll more smoothly in the text


orlinmalkja:
Nice idea!

orlinmalkja (approved):

PR #994, opened 2026-04-19
Implement target action areas

### LINKED ISSUE #993: Action Areas in Target Tooltip
When I click on a target in text and a tooltip appears, I want the tooltip to list all available annotations for this target, so that I get all the information in one place.

AC general

The tooltip will display action areas because there can be a mix of different annotation types belonging to that target. 
The action areas should offer their specific interactions: tooltip body actions, cross ref actions, sidebar annotation selection, etc. (YS: das könnte auch in EUPT hilfreich sein und inline Annotationen ersetzen) - see the attached image with a prototype.
There is no preview of annotations: only their labels.
 Clicking a nested target we show not only the direct annotations of it but also the annotations of its all parents.

AC for "tooltip body actions"

As tooltip annotations are usually very short: those annotations that have been configured to be shown as tooltips are displayed on top of the list and are fully visible. No further click is required to see a tooltip annotation.
The text of tooltip annotations can be selected and copied.
Tooltip annotations can contain hyperlinks.
No select

AC for "cross ref actions"

A cross-ref is listed second from the top and is labeled as "cross reference" or "Verweis".
Clicking on it activates cross-ref behaviour ("Open in this panel" "Open in a new panel")

AC for sidebar annotation selection

The corresponding annotation in side bar is selected (align, scroll to view)

Steps: Michelle macht die Daten für EUPT und BdN

Note: Long annotations automatically expand, bookmarking of tooltip annotations (share icon on hover)

orlinmalkja:
Closes #993


src/components/panel/TargetTooltipContainer.tsx
```diff
@@ -0,0 +1,70 @@
...
+        className="max-w-92 min-w-40"
+      >
+        <div
+          onClick={onClose}
+          className="absolute top-2 right-2 text-muted-foreground hover:text-foreground hover:cursor-pointer"
+          aria-label="Close"
+        >
+          <X size={16} />
+        </div>
+        <TargetTooltipContent crossRefAnnotation={crossRefAnnotation} relatedAnnotations={relatedAnnotations} onClose={onClose} />
```
paulpestov: Instead of prop drilling here, you can use children and use TargetTooltipContent directly in GenericTextRenderer

src/components/panel/TargetTooltipContent.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { usePanel } from '@/contexts/PanelContext.tsx'
+
+interface Props {
+  crossRefAnnotation: Annotation,
+  relatedAnnotations: Annotation[]
+  onClose: () => void
+}
+
+
+const TargetTooltipContent: FC<Props> = ({ crossRefAnnotation, relatedAnnotations, onClose }) => {
```
paulpestov: Was there a problem with passing an array of annotations? Why do you have to split it in crossRef and related?

src/components/panel/TargetTooltipContent.tsx
```diff
@@ -0,0 +1,67 @@
...
+        {tooltipAnnotations.map((ta) => <div className="mb-2"><TooltipAnnotation key={ta.id} annotation={ta} /></div>)}
+      </div>
+    )}
+    {crossRefInfo && <div>
+      {renderLabel(t('reference'))}
+      <CrossRefActionArea crossRefInfo={crossRefInfo} onSelect={handleCrossRefSelection} />
+    </div>}
+    {normalAnnotations.length > 0 && (
+      <div className={(crossRefInfo || tooltipAnnotations.length > 0) ? 'border-t-1 pt-2 border-muted-foreground' : ''}>
+        {renderLabel('Sidebar')}
```
paulpestov: translate

src/components/panel/TargetTooltipContent.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  function renderLabel(label: string) {
+    return <p className="px-2 py-1.5 text-sm font-medium">{label}</p>
+  }
+
+
+  return (<div className="w-[95%]">
+    {tooltipAnnotations.length > 0 && (
+      <div className={crossRefInfo ? 'border-b-1 pb-1 border-muted-foreground mb-2' : ''}>
+        {renderLabel('Tooltip')}
```
paulpestov: translate

src/components/panel/TargetTooltipContainer.tsx
orlinmalkja: yes, sounds good, changed it

src/components/panel/TargetTooltipContent.tsx
orlinmalkja: yes added it

src/components/panel/TargetTooltipContent.tsx
orlinmalkja: yes added it

paulpestov (approved):

PR #995, opened 2026-04-22
Add active style to tooltip target on click

### LINKED ISSUE #996: Add active style to tooltip target on click
ACs 
- clicking at a nested target should receive "expanded" style
- select at another nested target should add "expanded" style to the new target AND remove the "expanded" style of prev. sel. target
- closing the popover should remove the "expanded" style of target
- selecting a nested target should not affect the already selected annotation in sidebar AND add the "expanded" style to the new nested target

orlinmalkja:
Closes #996

### DIFF
--- a/src/components/GenericTextRenderer.tsx
+++ b/src/components/GenericTextRenderer.tsx
@@ -1,5 +1,6 @@
 import React, { FC, memo, useEffect, useRef, useState } from 'react'
 import {
+  addActiveTargetStyle,
   addAnnotationBaseStyle,
   addAnnotationId,
   addCrossRefTargetStyle,
@@ -17,7 +18,8 @@ import {
   getTargetsHoveredAnnotations,
   getTextTargets,
   isParentHovered,
-  isTargetPartOfSelectedAnnotation,
+  partOfSelectedTargets,
+  removeActiveTargetStyle,
   removeAnnotationBaseStyle,
   removeAnnotationIds,
   removeHighlightStyle,
@@ -78,6 +80,7 @@ const GenericTextRenderer: FC<Props> = memo(({
   const selectedAnnotationRef = useRef<Annotation | null>(null)
   const targetsRef = useRef<HTMLElement[]>(null)
   const hoveredAnnotationsRef = useRef<string[] | null>(null)
+  const clickedTargetRef = useRef<HTMLElement | null>(null)
 
   // Document object that is only recreated when htmlString changes - e.g. on item change or content type change
   const parsedDom: Element = React.useMemo(() => {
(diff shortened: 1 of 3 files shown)


src/components/GenericTextRenderer.tsx
```diff
@@ -364,6 +373,8 @@ const GenericTextRenderer: FC<Props> = memo(({
       }
     }
 
+    if (target !== clickedTargetRef.current)  addActiveTargetStyle(target)
```
paulpestov: I think this should happen inside of "if (openTooltip)" condition

src/components/GenericTextRenderer.tsx
```diff
@@ -270,9 +275,13 @@ const GenericTextRenderer: FC<Props> = memo(({
         someFiltered = !someFiltered ? fa.filtered[i] : true
       })
 
-      if (isTargetPartOfSelectedAnnotation(target, targetsOfSelectedAnnotation)) {
+      if (partOfSelectedTargets(target, targetsOfSelectedAnnotation)) {
         addSelectedStyle(target)
         return
+        // when click in target popover a sidebar annotation and deselect it again, then we land in this useffect where we recomputed the styles.
```
paulpestov: I disagree here because the active style should still be present since it was never removed. This condition should not be needed. 

Also adding active style here is an anti-pattern for our styling. You see above we are removing selected and hightlight styles pro-actively and add one of them later in the code. This is consistent since we remove/add the same styles. By just adding the active style you are creating a hanging responsibility to remove it which is not dealt with in this function.

paulpestov:
Bug: active style is applied when clicking an annotation without tooltip

src/components/GenericTextRenderer.tsx
```diff
@@ -78,6 +80,7 @@ const GenericTextRenderer: FC<Props> = memo(({
   const selectedAnnotationRef = useRef<Annotation | null>(null)
   const targetsRef = useRef<HTMLElement[]>(null)
   const hoveredAnnotationsRef = useRef<string[] | null>(null)
+  const clickedTargetRef = useRef<HTMLElement | null>(null)
```
paulpestov: Rename to "activeTargetRef"

paulpestov:
In general I think it would be ok to shift a bit away from our styling pattern when it comes to active target. Normally, we try to append only one type of styling: highlight or hover or nested or selected. 

In case of active I think we can work with overwriting in order to keep the old style still present and not try to solve it in useEffects. So maybe it would be good to add "important" to the active style and just add/remove it on click/close and leave the rest alone.

orlinmalkja:
@paulpestov I agree with all your points above and fixed your suggestions! For testing purposes I changed the active style to yellow color. I myself missed i.e the above bug due to the same style as selected :D

src/components/GenericTextRenderer.tsx
orlinmalkja: yes

src/components/GenericTextRenderer.tsx
orlinmalkja: I agree, changed it

orlinmalkja:
> Bug: active style is applied when clicking an annotation without tooltip

should be fixed now

src/components/GenericTextRenderer.tsx
orlinmalkja: changed it

orlinmalkja:
> In general I think it would be ok to shift a bit away from our styling pattern when it comes to active target. Normally, we try to append only one type of styling: highlight or hover or nested or selected.
> 
> In case of active I think we can work with overwriting in order to keep the old style still present and not try to solve it in useEffects. So maybe it would be good to add "important" to the active style and just add/remove it on click/close and leave the rest alone.

I find this a good idea and this is kind of 'separation of concerns' in software development ?

paulpestov (approved):

PR #1005, opened 2026-04-26
feat: use "selectedIndex" in annotation filters config to specify a selected filters tab instead of a "selected" option at each filter node

paulpestov:
- look at new bdn config
- look for "selectedIndex" key
- look in README for explanation

behaviour: ony selectedIndex is allowed to set at config to control selected state of filters, everything else is considered as "selected"

in code: annotationFilters in PanelContext now represents the current state of filters, therefore a new interface FilterNodeWithSelection was introduced. It adds "selected" key to FilterNode while FilterNode stays without it in order to be exported to the users. Config is only used when needed, e.g. preset annotationFilters at panel init or decide whether to render SingleRootFilter or MultiRootFilter component.


paulpestov:
@orlinmalkja pls review now :)

orlinmalkja (approved):

PR #1010, opened 2026-04-28
Sort target popover annotations based on deepest clicked target

### LINKED ISSUE #1009: Sort the tooltip and sidebar annoations in Annotation Popover
In Annotation Popover we display target's tooltip and sidebar annotations.
Based on a requirement we want to sort for each group (tooltip and sidebar) the annotations so that we first list the annotations to deepest clicked target and further below all other annotations.

orlinmalkja:
Closes #1009 

I think we can sort the tooltip annotations and then display the content. Therefore we just to updatea boolean variable as state through `setDisplayAnnotations` when we finish sorting the annotations

### DIFF
--- a/src/components/panel/annotations/popover/AnnotationPopoverContent.tsx
+++ b/src/components/panel/annotations/popover/AnnotationPopoverContent.tsx
@@ -1,4 +1,4 @@
-import { FC, useEffect, useState } from 'react'
+import { FC, useEffect, useRef, useState } from 'react'
 import { useConfig } from '@/contexts/ConfigContext.tsx'
 import { getCrossRefInfo } from '@/utils/annotations.ts'
 import TooltipItem from '@/components/panel/annotations/popover/items/TooltipItem.tsx'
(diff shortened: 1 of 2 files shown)


src/components/panel/annotations/popover/AnnotationPopoverContent.tsx
```diff
@@ -7,55 +7,78 @@ import BaseItem from '@/components/panel/annotations/popover/items/BaseItem.tsx'
...
+  }
+
 
   function renderLabel(label: string) {
     return <p className="py-1 mb-1 text-xs font-medium text-muted-foreground">{label}</p>
   }
 
   return <div className="flex flex-col gap-4">
-    {tooltipAnnotations.length > 0 && (
+    {tooltipAnnotationsRef.current?.length > 0 && (
```
paulpestov, line 69: I find it very strange that you often use refs for template rendering. Can you elaborate on the purpose? State is exactly meant to be used in a template, so the template can be kept up-to-date. This is the whole idea behind React. If you have any trouble to update the tempalte accordingly, it tells that there is a problem in your logic which should be fixed rather than using a ref.

src/components/panel/annotations/popover/AnnotationPopoverContent.tsx
```diff
@@ -7,55 +7,78 @@ import BaseItem from '@/components/panel/annotations/popover/items/BaseItem.tsx'
...
   const [crossRefInfos, setCrossRefInfos] = useState<CrossRefInfo[]>([])
-  const tooltipAnnotations = relatedAnnotations.filter(a => tooltipTypes.includes((a.body as AnnotationBody)['x-content-type']))
-  const normalAnnotations = relatedAnnotations.filter(a => !tooltipTypes.includes((a.body as AnnotationBody)['x-content-type']))
-
-  function handleCrossRefSelection() {
-    setCrossRefInfos([])
-    onClose()
-  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [_, setDisplayAnnotations] = useState(null)
```
paulpestov: Use loading instead of this

src/components/panel/annotations/popover/AnnotationPopoverContent.tsx
orlinmalkja: I am not sure whether I am using ref that often for template rendering. As far as I remember I have used it mostly in Cross Ref feature (the first implementation and now). 

I thought to use ref for the following reason: Why to have two states for 'tooltipAnnotations' and 'normalAnnotations', when on target click they are computed and just rendered in Popover. So I think we just need one boolean to check if the computation of these values finished. Once it finished then we render them.  Secondly, whenever we open the popover then the data (tooltipAnnotations and normalAnnotations) do not change anymore. So there is no further UI interaction within the Popover which might change them. 

No, I did not have problems on updating the template - this was just an optimisation step, not to rerender twice the Popover for each 'setNormalAnnotations', 'setTooltipAnnotations'.

src/components/panel/annotations/popover/AnnotationPopoverContent.tsx
orlinmalkja: yes

paulpestov (approved):

PR #1011, opened 2026-04-28
Bug clicking on cross ref link nothing is shown

### LINKED ISSUE #1012: Bug: clicking in cross ref link nothing is shown
**Describe the bug**

**To Reproduce**
1. Go to '..bdn.json', griesbach, digitale ansicht, Vorrede
2. Click on link 103 as in the screenshot below

Image
 <br/><br/>

**Actual behavior**
As we can see below. A very small popover is shown with no text on it. The user does not know why he can't jump with the link to the required destination.
An error is in console. <br/> <br/>

Image  <br/> 

**Expected behavior**
When there exists errors regarding the CrossRef then the user should be able to see them.

orlinmalkja:
Closes #1012 

The crossRefData is not correctly given, due to having in source.collection a root Collection url and not a leaf collection. As a consequence the manifest and item data are not valid data structure.

In `annotations.ts` we have function getCrossRefInfo. When refItemData is wrong data, i.e does not have 'content' attribute, then the code should not break and return the retrieved data from crossRefAnnotation.
We already handle errors in crossRef Data in `CrossRefItem`.
By making this small fix then we would see the content as in following screenshot <br/> <br/>
image

Additionally I just restore the validation function of the existence of target in referenced html text

### DIFF
--- a/src/components/panel/annotations/popover/items/CrossRefItem.tsx
+++ b/src/components/panel/annotations/popover/items/CrossRefItem.tsx
@@ -3,6 +3,7 @@ import { usePanel } from '@/contexts/PanelContext.tsx'
 import { validateCrossRefNode } from '@/utils/cross-ref.ts'
 import { CustomError } from '@/utils/custom-error.ts'
 import CrossRefLink from '@/components/panel/annotations/popover/cross-ref/CrossRefLink.tsx'
+import { existsTargetInText } from '@/utils/dom.ts'
 
 interface Props {
   crossRefInfo: CrossRefInfo,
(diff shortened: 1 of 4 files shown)


src/components/panel/annotations/popover/items/CrossRefItem.tsx
```diff
@@ -30,8 +31,9 @@ const CrossRefItem: FC<Props> = ({ crossRefInfo, onSelect }) => {
             manifestLabel: (manifestData as Manifest).label,
             itemLabel: newItemLabel
           }
-          // TODO: if (!await existsTargetInText(extendedCrossRefInfoRef.current.refItemData, extendedCrossRefInfoRef.current.contentType, extendedCrossRefInfoRef.current.selector)) throw new CustomError('cross_ref_error_title', 'referenced_element_not_found')
-          loadedData.current = true
+          const targetExists = await existsTargetInText(extendedCrossRefInfoRef.current.refItemData, extendedCrossRefInfoRef.current.contentType, extendedCrossRefInfoRef.current.selector)
+          if (!targetExists) setError(new CustomError(t('cross_ref_error_title'), t('referenced_element_not_found')))
+          else loadedData.current = true
```
paulpestov: Why is this needed?

orlinmalkja: For cross ref referring to text part. Throw error when referenced el cannot be found due to a false provided selector. 
In this case we show the error, i.e the target el could not be found due to a wrong selector. This is informative and can help fix the selector value. The negative point is that the user does not jump to the referenced item. 

I think a better approach would be to allow the user to jump to the referenced item and then there to show a kind of toast error that the ref.target selector is provided wrong.

paulpestov (approved):

PR #1017, opened 2026-05-07
feat!: upgrade TextAPI support to v2.0.0

paulpestov:
Migrates the app to support TextAPi 2.0. Key changes:

- Types are now split into 2 files, textapi.d.ts contains (manually) generated types and interfaces from the TextAPI openapi.json
- Check those types to understand how they affect the app
- A change that caused lots of updates is the fact that there are no "sequences" anymore. Instead there are specific keys for previously known sequence items ("collections", "manifests", "items"). Additionally, those new sequences can also be string[]. So we need to check everytime whether it is a nested object or a string.
- AnnotationAPI is not migrated yet.


orlinmalkja:
Thats a really large work ! Thanks for that! 
Here is the first bug I found: Overlapping of annotations in align mode

image

orlinmalkja:
Bug 2: The list of annotations does not correspond completely to the highlighted targets
We see one target, but a list of 6 annotations

image

orlinmalkja:
Bug 3: Using `example.json`: the list of annotations does not reflect that from selected filters. By considering the selected filters, then we should not display any of below annotations, right ?

image

image

orlinmalkja:
Bug: When navigate to a new item - We should get the content types of the new item. We should check if previous active content type belongs to content types of new item. If this is the case, then we should preserve the active content type. If not, then the new active content type of the item becomes the first one.

orlinmalkja:
Small bug: Set correctly label names of each child node in tree (starting from second level) - please see the screenshot

image

paulpestov:
> Bug: When navigate to a new item - We should get the content types of the new item. We should check if previous active content type belongs to content types of new item. If this is the case, then we should preserve the active content type. If not, then the new active content type of the item becomes the first one.

fixed

orlinmalkja:
Bug: should not disable 'prev-item' button when there are previous items

image

orlinmalkja:
@paulpestov I finished the testing with `Vier Wachen` data. That's all I found. Next I'll try to test it with `Melchior Hoffmann` data.

paulpestov:
> Bug: should not disable 'prev-item' button when there are previous items

fixed

orlinmalkja (approved):

Issue #1013, opened 2026-05-04
Bug: without filter config - CrossRef is shown in filters


orlinmalkja:
- Specifying no filters in config, TIDO should display the annotations types by reading them from TextAPI content
- TIDO should not include in annotation types filtering the 'CrossRef'

PR #1021, opened 2026-05-12
fix: handle click event on target without filtered annotations

### LINKED ISSUE #1015: Filtered-out annotations are still clickable in text and visible in popover
**Describe the bug**
When we use filter to show only certain annotation types (ex: "Textkritik") filter works for the sidebar list - but text itself and popover doesn't reflect it. So I can still click and see annotations that I filtered out.

**To Reproduce**
- Go to `Digitale Ansicht - Vorrede(n)`
- Open the filter and switch to a different group tab (ex: from `Textkritik` to `Erläuterungen`)
- Sidebar correctly shows only the active group's annotations
- But, clicking on a word in text still shows annotations from other (inactive) group in the Popover (ex: annotation icon with speech bubble in text)
- Annotations text that only belong to the inactive group are still clickable in the text (ex: when Erläuterungen group is active and click on text)

**Expected behavior**
Only annotations from currently active filter group should be visible and clickable - in the sidebar, popover, and also in the text. Inactive groups should be fully hidden everywhere.

orlinmalkja:
- prevent click on this target
- show the correct number of filtered annotations which belong to target

How I tested this PR: Well I used another branch `fix-scroll-selected-annotation` which does not include the newest PR which adapts TIDO with TextAPI 2.0. 
As data for bdn I use the current data, which is not yet migrated to 2.0 

Closes #1015

### DIFF
--- a/src/components/panel/renderers/text/GenericTextRenderer.tsx
+++ b/src/components/panel/renderers/text/GenericTextRenderer.tsx
@@ -82,6 +82,7 @@ const GenericTextRenderer: FC<Props> = memo(({
   const targetsRef = useRef<HTMLElement[]>(null)
   const hoveredAnnotationsRef = useRef<string[] | null>(null)
   const activeTargetRef = useRef<HTMLElement | null>(null)
+  const selectedAnnotationTypesRef = useRef<AnnotationTypesDict | null>(null)
 
   // Document object that is only recreated when htmlString changes - e.g. on item change or content type change
   const parsedDom: Element = React.useMemo(() => {
@@ -296,7 +297,7 @@ const GenericTextRenderer: FC<Props> = memo(({
 
     setMatchedMap(resultMap)
     if (onUpdateMatchedAnnotationsMap) onUpdateMatchedAnnotationsMap(resultMap)
-
+    selectedAnnotationTypesRef.current = selectedAnnotationTypes
   }, [selectedAnnotationTypes])
 
   const onMouseEnterTarget = (e: Event) => {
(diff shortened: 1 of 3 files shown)


src/components/panel/renderers/text/GenericTextRenderer.tsx
```diff
@@ -317,6 +318,18 @@ const GenericTextRenderer: FC<Props> = memo(({
     setHoveredAnnotations(hoveredAnnotationsRef.current?.filter(a => !idsArray.includes(a)) ?? null)
   }
 
+  function isFilteredAnnotation(annotation: Annotation, selectedAnnotationTypes: AnnotationTypesDict) {
+    // filter Variant Annotations based on witnesses in selectedAnnotationTypes
+    // filter all other annotations which have type as key in selectedAnnotation types
+    const annotationType = annotation.body['x-content-type']
```
paulpestov: x-content-type does not exist anymore, use annotationType key

src/components/panel/renderers/text/GenericTextRenderer.tsx
```diff
@@ -343,21 +363,20 @@ const GenericTextRenderer: FC<Props> = memo(({
         return Array.from(parsedDom.querySelectorAll(selector)).includes(target)
       })
 
-    const crossRefContentType = annotationsConfig?.crossRefContentType
     const seen = new Set<string>()
     // compute related annotations: all annotations for the clicked target and its parent targets
     const newRelatedAnnotations = (flippedMatchedMapRef.current ?? [])
       .filter(entry => entry.target === target || entry.target.contains(target as HTMLElement))
       .flatMap(entry => entry.annotations)
-      .filter(a => !seen.has(a.id) && seen.add(a.id) && getAnnotationContentType(a) !== crossRefContentType)
```
paulpestov, line 352: why was this replaced? not used anymore?

orlinmalkja: because I am filtering annotations based on filters selection. So I created a function isFilteredAnnotation. It checks for variant annotations as well. 

Now we have a unified format of annotation body which has inside it the `annotationType` key right ? Then I think we do not need getAnnotationContentType anymore

src/components/panel/renderers/text/GenericTextRenderer.tsx
orlinmalkja: okay

src/components/panel/renderers/text/GenericTextRenderer.tsx
paulpestov: Yes I can see this, but I asked for the "why". Checking for the crossRefContentType was previously necessary somehow. Why is it not anymore? You know there is also this bug #1013, does this line also have influence?

orlinmalkja: Well, crossRefContentType cannot appear in selectedAnnotationTypes right ? If not, then we already filter the right types by using       return Object.keys(selectedAnnotationTypes).includes(annotationType). 
Therefore, we do not include crossRefAnnotations as filteredAnnotations

paulpestov: I hope you are right :)

paulpestov (approved):

PR #1037, opened 2026-05-20
refactor: use paddingTop prop directly as value in GenericTextRenderer

### LINKED ISSUE #1036: Bug: in example data Annotation is cropped in 3 lines not nicely
Image

orlinmalkja:
- Refactor the prop `paddingTop` in GenericTextRenderer - now it is a number and not extract it implicitly from another variable such as `showContentTypeToggle`
- this fixes the cropped annotation in 3 lines, which looks badly cropped

Closes #1036


orlinmalkja:
@paulpestov merging this

PR #1046, opened 2026-05-24
Implement panel fullscreen mode

### LINKED ISSUE #1047: Clicking at full screen in panel should open Panel in full screen
- Clicking fullscreen button in panel should open it in Full Screen
- It should be possible to exit fullscreen by clicking the button or typing ESC key
- In fullscreen - opening annotations sidebar should accomodate place for that
- Clicking at any label in Panel should open respective dropdown, i.e in CollectionTItle, Manifest, Item Label, PanelViews

orlinmalkja:
This PR adds following
- Clicking fullscreen button in panel should open it in Full Screen
- It should be possible to exit fullscreen by clicking the button or typing ESC key
- In fullscreen - opening annotations sidebar should accomodate place for that
- Clicking at any label in Panel should open respective dropdown, i.e in CollectionTItle, Manifest, Item Label, PanelViews

Approach: We use requestFullScreen browser function to open a certain html element in full screen. The problem with this is that the Popover, Radix elements since they are append to body by default, will not be shown. Thus I create a ref - container ref in the PanelContext. Each component which displays such Popover, Radix then uses this container to portal them into the container ref, in order to be able to see them in full screen. 

Closes #1047

### DIFF
--- a/public/translations/de.json
+++ b/public/translations/de.json
@@ -75,6 +75,8 @@
     "please_select_panel_mode": "Bitte wählen Sie einen Modus",
     "show_image": "Zum Bild",
     "show_text": "Zum Text",
+    "show_fullscreen": "Im Vollbildmodus anzeigen",
+    "exit_fullscreen": "Vollbildmodus beenden",
     "jump_to": "Springen zu",
     "reference": "Verweis",
     "text_type": "Texttyp",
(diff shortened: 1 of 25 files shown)


paulpestov:
This feature lacks a concept for small panel sizes, please ensure that the panel header does not break when a panel has it's minimal width. Maybe we need to introduce the "..." menu again.

PR #1057, opened 2026-06-04
Fix cross ref in annotations

### LINKED ISSUE #1059: Bug: Cannot click Cross Ref in Annotation

orlinmalkja:
Closes #1059, #1060, #1061

This PR adds following
- Fixes the bug of clicking a cross ref in annotation
- Adds a source as prop in BaseItem in annotations - we can't continue with activeContentUrl, since the annotation popover can be opened in TextRenderer or in Annotation.tsx
- Report and handle error better when facing such in getCrossRefInfo
- add example data of cross ref in annotations which refer to text or in annotation

I find it weird that in github pipeline now with node version 22 it passes all test successfully, while with 24 not. However locally all tests pass

### DIFF
--- a/src/components/panel/annotations/popover/AnnotationPopoverContent.tsx
+++ b/src/components/panel/annotations/popover/AnnotationPopoverContent.tsx
@@ -1,5 +1,6 @@
 import { FC, useEffect, useRef, useState } from 'react'
 import { getCrossRefInfos } from '@/utils/annotations.ts'
+import { CustomError } from '@/utils/custom-error.ts'
 import TooltipItem from '@/components/panel/annotations/popover/items/TooltipItem.tsx'
 import CrossRefItem from '@/components/panel/annotations/popover/items/CrossRefItem.tsx'
 import BaseItem from '@/components/panel/annotations/popover/items/BaseItem.tsx'
@@ -25,6 +26,7 @@ const AnnotationPopoverContent: FC<Props> = ({
   const { t } = usePanelTranslation()
 
   const [crossRefInfos, setCrossRefInfos] = useState<CrossRefInfo[]>([])
+  const [crossRefErrors, setCrossRefErrors] = useState<CustomError[]>([])
   const [loading, setLoading] = useState(true)
 
   const tooltipAnnotationsRef = useRef<Annotation[]>(null)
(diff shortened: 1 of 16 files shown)


src/components/panel/annotations/popover/AnnotationPopoverContent.tsx
```diff
@@ -43,8 +47,10 @@ const AnnotationPopoverContent: FC<Props> = ({
     }
 
     async function computeCrossRefInfos(annotations: Annotation[]) {
-      const infos: CrossRefInfo[] = await Promise.all(annotations.map(a => getCrossRefInfo(a))) as CrossRefInfo[]
-      setCrossRefInfos(infos)
+      const results = await getCrossRefInfos(annotations)
+
+      setCrossRefInfos(results.filter((r): r is CrossRefInfo => !(Object.hasOwn(r, 'name') && Object.hasOwn(r, 'message'))))
```
paulpestov: As discussed yesterday, try to solve this problem within CrossRefItem component by passing the annotation as prop. Let React decide when to render and when to show errors. This does not need to happen at once with "allSettled" or "all". 

Generally speaking I think it is a good practice to pass annotation to indivudual Popover list elements as it provides a common and consistent interface. Then the items decide what to do and how to render that annotation.

orlinmalkja: thank you for your feedback! yes I did that.

orlinmalkja:
@paulpestov I checked the test in panel.cy regarding "Should navigate in local tree". It actually checks for 'loading panel' first is true, that means the new item is being loaded. And then it checks for 'Loading panel' is false -> new text is loaded. That means we test this correctly.

At some point today all tests passed and then 2 failed afterwards 1 failed. I think the tests in github pipeline or github pipeline does not seem stable to me ... In contrary all tests locally pass successfully

Please feel free to test again cross ref in annotations and in text. I think my contribution for this PR is enough. Some further improvement I would like to make in a new one.

src/utils/annotations.ts
```diff
@@ -143,6 +157,16 @@ async function getCrossRefInfo(annotation: Annotation): Promise<CrossRefInfo> {
   }
 }
 
+async function getCrossRefInfos(annotations: Annotation[]): Promise<(CrossRefInfo | CustomError)[]> {
```
paulpestov: Unused?

orlinmalkja: @paulpestov yes, removed it

PR #1080, opened 2026-06-24
Implement panel synopsis

orlinmalkja:
- Find SyncTargets of clicked target on demand by using a new data structure syncAnnotationsBySource. Reason for this is to avoid syncMapRef and some other previous functions i.e updating this ref when new panels are opened which might introduce new sync annotations.


orlinmalkja:
@paulpestov Here I change the approach instead of syncMapRef to computing the syncTargets on demand by click. By doing so the code becomes much cleaner and we do not include some additional functions that we would need to add in the approach with syncMapRef. Indexing the syncAnnotations by contentUrl should help in improving the performance of this finding. What do you think of this new approach to get rid of syncMapRef and compute the syncedTarget on demand by click ?

src/components/panel/renderers/text/GenericTextRenderer.tsx
```diff
@@ -34,11 +36,48 @@ import {
...
+import { useShallow } from 'zustand/react/shallow'
 import { useConfig } from '@/contexts/ConfigContext.tsx'
 import AnnotationPopoverContainer from '@/components/panel/annotations/popover/AnnotationPopoverContainer.tsx'
 import AnnotationPopoverContent from '@/components/panel/annotations/popover/AnnotationPopoverContent.tsx'
+import SynopsisItem from '@/components/panel/annotations/popover/items/SynopsisItem.tsx'
 import { SelectedAnnotation } from '@/types'
 
+
+// Only CssSelectors carry a `value`. RangeSelectors are not handled yet (see findTargets in utils/annotations).
+function getSelectorValue(target: AnnotationTarget): string | null {
```
paulpestov: move to utils

paulpestov:
Did you test it with pane synopsis annotations present? Maybe add a few into example data, also some targets that are both: pane and panel synoptical.

src/components/panel/renderers/text/GenericTextRenderer.tsx
```diff
@@ -154,34 +239,31 @@ const GenericTextRenderer: FC<Props> = memo(({
...
+  useEffect(() => {
+    if (!parsedDom || sourceSyncAnnotations.length === 0) return
+
+    sourceSyncAnnotations.forEach((cur) => {
+      const target = cur.target.find((t) => getSource(t).id === source)
+      const selector = target ? getSelectorValue(target) : null
+      if (!selector) return
+
+      Array.from(parsedDom.querySelectorAll(selector)).forEach((el) => {
+        if (getSyncAnnotationIds(el).some(Boolean)) return
```
paulpestov, line 296: I would guess that this can lead to problems where the target is from pane and panel synopsis at the same time, couldn't it?

src/components/panel/renderers/text/GenericTextRenderer.tsx
```diff
@@ -404,14 +503,16 @@ const GenericTextRenderer: FC<Props> = memo(({
...
-    const openTooltip = _tooltipAnnotations.length > 0 || crossRefAnnotations.length > 0 || normalAnnotations.length > 1
+    const openTooltip = _tooltipAnnotations.length > 0 || crossRefAnnotations.length > 0 || normalAnnotations.length > 1 || newSyncTargets.length > 0
 
     if (openTooltip) {
       setTooltipOpen(true)
       setTooltipTargetElement(target as HTMLElement)
       setRelatedAnnotations(normalAnnotations)
       setTooltipAnnotations(_tooltipAnnotations)
       addActiveTargetStyle(target)
+      // 2) pass the synced targets of this entry (and the clicked target's y-position) to the popover content
```
paulpestov: where is 1)^^

src/components/panel/annotations/popover/items/SynopsisItem.tsx
```diff
@@ -0,0 +1,125 @@
...
+        openInNewPanel(source)
+        return
+      }
+
+      // 2) source.item is already opened in another panel
+      // 2b) it already shows the text view for source.id -> do nothing
+      if (panelShowsSource(otherPanel, source)) return
+
+      // 2a) it misses the text view for source.id -> add one
+      usePanelStore.getState().updatePanel(otherPanel.id, {
```
paulpestov: This is nice, but it violates the config. We have to mention this in MEKONG, since there is no way back to the configured state.

paulpestov:
> @paulpestov Here I change the approach instead of syncMapRef to computing the syncTargets on demand by click. By doing so the code becomes much cleaner and we do not include some additional functions that we would need to add in the approach with syncMapRef. Indexing the syncAnnotations by contentUrl should help in improving the performance of this finding. What do you think of this new approach to get rid of syncMapRef and compute the syncedTarget on demand by click ?

Not sure what you mean, I never saw a syncMapRef. I get the concept ofc but never saw it in action. The current state isn't really "on-demand", isn't it? It just selects from the store by source, but  the targets are being computed at init time, right? Anyways this works for me and is a good first implemention. There is some potential for refactoring, though. Especially in the GenericTextRenderer which becomes more and more a trash bin for wild calculations. Adding SynopticItem as child to Popover does not pay in towards SOC or reusability. However, we knew before this feature that the GTR needs refactoring, so its ok for now.

src/components/panel/renderers/text/GenericTextRenderer.tsx
orlinmalkja: yes did it

orlinmalkja:
> Did you test it with pane synopsis annotations present? Maybe add a few into example data, also some targets that are both: pane and panel synoptical.

@paulpestov 
yes I tested it and added example data for this purpose: open example-synopsis-2, manifest 2, item 1. Click on Ishmael. It has 6 targets (among them one targets in the same panel and others in different items). 

I have a follow up question: When a new target i.e lies in 'normalized' of same item. Do we append this pane to the current panel we clicked the target or in a new panel. I think that Vier Wachen wants to open in a new panel... I don't remember this exactly. 
When the user has configured 2 views (text translation, transcription), should we append a third view ?

When thinking about this, when the new text belongs to same item, it can be opened in same panel no ? - I think it is not so consistent having by default two text of same item in same panel and then open a third text of same item in a new panel , is it ?

paulpestov:
> Do we append this pane to the current panel we clicked the target or in a new panel.

I think I wrote a comment about this above, we should ask MEKONG

paulpestov (approved):

orlinmalkja:
In beta 4 release the pipeline failed (with 11 tests). Pipeline here fails the same as well. Locally all tests pass, thus I am merging.

PR #1091, opened 2026-07-02
Add panel synopsis witness selection

### LINKED ISSUE #1088: Panel Synopsis Witnesses Selection in Popover
ACs
- In the Popover we should be able to select which witnesses (manifests) to open
- Should display each witness containing a synoptic target as a check box. If the witness is already opened in any panel its checkbox should be preselected
- Should display an 'all' checkbox. Clicking it should select all witnesses checkboxes
- Should display a 'replace' checkbox/button. Selecting replace should replace existing panels (except the current clicked one) with the panels of selected witnesses. 
   - replace scenario: The current panel should be shown to the left-most and the new panels should open to the right of it
- Should display a button 'Open in panels'. Clicking at this button open the desired witnesses and closes the popover. This button should be initially disabled/muted. Once we have at least one witness selected, then it becomes active

orlinmalkja:
Closes #1088

### DIFF
--- a/public/translations/de.json
+++ b/public/translations/de.json
@@ -96,6 +96,10 @@
     "witnesses": "Textzeugen",
     "selected_witnesses": "{{selected}} von {{total}} Textzeugen ausgewählt",
     "unknown_witness": "Unbekannter Textzeuge",
+    "synoptical_witnesses": "Synoptische Zeugen",
+    "select_all": "Alle auswählen",
+    "witnesses_count": "{{selected}} von {{total}}",
+    "witness_open": "offen",
     "options": "Optionen",
     "swap": "Swap",
     "split": "Split",
@@ -140,6 +144,8 @@
     "show_in_sidebar": "In Seiteleiste anzeigen",
     "more_annotations": "Weitere Annotationen",
     "synopsis": "Synopse",
-    "open_synced_panels": "In Panels öffnen"
+    "open_synced_panels": "In Panels öffnen",
+    "replace_panels": "Bestehende Panels ersetzen",
+    "replace_panels_hint": "Andere Panels werden geschlossen. Das aktuelle Panel bleibt links, neue öffnen rechts davon."
   }
 }
--- a/public/translations/en.json
+++ b/public/translations/en.json
@@ -91,6 +91,10 @@
     "witnesses": "Witnesses",
     "selected_witnesses": "{{selected}} of {{total}} witnesses selected",
     "unknown_witness": "Unknown witness",
+    "synoptical_witnesses": "Synoptical Witnesses",
+    "select_all": "Select all",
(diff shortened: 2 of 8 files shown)


paulpestov:
The `select all` checkbox does not show the correct state. It should switch to "checked" when all are selected.

paulpestov:
Clicking `replace existing panels` toggles a description text, so the "Open" button moves down -> eventually out of screen and becomes unreachable. Suggestion: Show the description text in a tooltip.

paulpestov:
The number of selected witnesses does not update correctly if I see correctly. When selecting "Ishmael" and checking all 6 checkboxes, it stays at "4 of 6". Also there are 2 displays of this, one next to "selected witnesses" and one next to "select all". Go only with "selected witnesses".

paulpestov:
I am a little overwhelmed with the witness items. I think it is too much cognitive load for the users. Even though it looks nice, it gives me headache about the meaning of each entry. There is too much information in one single entry. 

- Checkbox - the user has to understand what "checking" will do -> look down and see "Open new panels" button
- Chip - Manifest/Siglum -> compare this with titles in the panel to check whether they are the same
- Manifest title - additional Info, could be useful or could be hidden as either siglum or this title is enough to continue
- Open badge - One of the first things jumps into sight, good but how does it work together with the checkbox?

These are things that the users have to worry about for one single item and if there are 6 of them, it multiples the cognitive load.

My suggestion would be to drastically simplify the entries: 
https://github.com/subugoe/tido/issues/1088#issuecomment-4868531450


In order to continue with this PR, you could also try it out your suggestion but maybe merge same witnesses into one entry. Or maybe just the example we have is it far more complicated than the real data would be and your suggestion will work out. Then you could get feedback and if it is desired to simplify, you could try some UI elements from my mockup.

orlinmalkja:
> The `select all` checkbox does not show the correct state. It should switch to "checked" when all are selected.

The current solution uses an object: with `manifestId`: `boolean`. As we talked today in digital humanities each witness is a different manifest. In our example data we have sync targets which belong to same manifest for the same clicked target. In our object for those sync targets they will be added only once in the object. Therefore select all does not work all

Having sync targets belonging to the same manifest is not a realistic scenario in digital editions. However I fixed this by identifying as opened the sync target whose source is opened in one panel. As key I used 'source.id' = 'contentUrl'. That means a witness obtains 'open' badge if the sync target's exact text is opened in any panel.

orlinmalkja:
> The number of selected witnesses does not update correctly if I see correctly. When selecting "Ishmael" and checking all 6 checkboxes, it stays at "4 of 6". Also there are 2 displays of this, one next to "selected witnesses" and one next to "select all". Go only with "selected witnesses".

The reason for this bug is the same as I described above. Regarding the duplicates of showing "4/6" in two places, yes sounds good, let's keep only the "selected witnesses" one.

